PR #10, opened 2017-09-22
Update on the assumption that this is public already

### LINKED ISSUE #9: Creating an issue will ultimately only require 'public_repo'
We should probably tweak the help so that this makes sense when the repo becomes public.

martinthomson:
Closes #9

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1,2 +1,13 @@
 [
+  {
+    "ciuName": null, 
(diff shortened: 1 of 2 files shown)


hildjj:
Now has conflicts.  please resolve.

dbaron:
I merged the activities.py change manually.

PR #46, opened 2017-11-05
Changes needed to process http://iot.mozilla.org/wot/

dbaron:
These are the changes I needed to process the Web Things API spec.  In particular:
* there's a title but not an h1 at the top, since the h1 gets generated dynamically by respec's JS
* the element with id=abstract is a section rather than an h2/h3
* the spec's URL wasn't in URL2ORG

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -346,11 +346,16 @@ class W3CParser(SpecParser):
         try:
             data['title'] = spec.h1.string
         except AttributeError:
-            sys.stderr.write("* Can't find the specification's title.\n")
-            sys.exit(1)
+            try:
+                data['title'] = spec.title.string
+            except AttributeError:
+                sys.stderr.write("* Can't find the specification's title.\n")
+                sys.exit(1)
         try:
-            data['description'] = self.clean_tag(
-                spec.find(id='abstract').find_next_sibling(["p", "div"]))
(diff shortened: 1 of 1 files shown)


activities.py
```diff
@@ -447,6 +452,7 @@ def html_url(doc_name):
     'wicg.github.io': W3CParser,
     'dev.w3.org': W3CParser,
     'dvcs.w3.org': W3CParser,
+    'iot.mozilla.org': W3CParser,
```
martinthomson: Check with @benfrancis, but I think that this is a one-off site and maybe not worth adding to this list.  Maybe a `-p` command-line option would be a better way to drive the selection of the parser for this sort of one-off.

benfrancis: I'm hoping that the draft member submission at [iot.mozilla.org/wot](http://iot.mozilla.org/wot) will eventually be superseded by a W3C specification. That work is starting with [this Interest Group Note](http://w3c.github.io/wot/proposals/json-td/). But so far that only covers one area of our draft member submission, so we'll have to see how that work progresses. In the meantime we may continue to add to the draft member submission to reflect our latest thinking in its completeness.

dbaron:
I dropped the change to URL2ORG, and I'm going to just go ahead and merge the rest.

Issue #51, opened 2017-11-20
Set of possible positions seems too narrow


annevk:
At least for PRs against WHATWG standards we need a stronger signal than "participating" gives. We need something that indicates support. That's not a commitment to implement, but it is an endorsement of the feature that allows others to go ahead and implement it, to add tests to web-platform-tests, etc.

My initial hope for this repository was that we could have something where one could raise a standards-related issue with Mozilla stakeholders and determine a position on that issue. The way it's structured now doesn't really lend itself to that.

dbaron:
Some prior history here is in #16 and #41, although a bunch of it was in email threads.  I'd be happy to get rid of `participating`, which I think is a statement of fact rather than a position.  If I were replacing it I'd probably add `important` and `neutral`.

tantek:
I agree with re-emphasizing @annevk's use-case of "determine a [Mozilla] position on that issue". Also agreed that "participating" is more of a statement of fact (whether in WG membership, spec editorship, or spec issues/pull requests).

I've also re-reviewed the #16 and #41 issues and related threads for nuances we may have overlooked.

> I'd probably add important and neutral.

I think "important" is too vague (relative to what?), and "neutral" doesn't say anything (enough?)

Counterproposal: add non-harmful, worth-prototyping

I believe this keeps with having mutually exclusive positions in rough time / positivity order:

We have no position yet:
* under consideration - (but we're looking at it)
* defer - (we have postponed looking at it - not looking at it currently)

We have a position:
* harmful - Mozilla considers this specification to be harmful in its current state.
* non-harmful - (we've looked at it enough to determine that this spec is likely not harmful in its current state, but we're not sure of actual utility vs cost to implement, or whether this is a good enough approach to the problem space etc.)
* worth-prototyping - this indicates conceptual support (more than not harmful, we think the spec is at least good enough to be worthy of implementers prototyping to incubate/iterate, but not saying whether we are or are not. Actual state of whether *we* prototype/implement is separately covered by Intent to Experiment, Intent to Implement, Intent to Ship etc.) we may actually prototype, and then find there are bad enough problems to change state to "harmful" for example. or we may find that the spec is awkward/subpotimal but not actually harmful and change state to "non-harmful".

Feel free to bikeshed the names, but those two conceptual additions (non-harming, worth-prototyping), I believe provide sufficient additional granularity to weakly or strongly positively communicate our positions.

[...]

Issue #29, opened 2017-09-27
RFP: Web Packaging Format


marcoscaceres:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Web Packaging Format
* Specification or proposal URL: https://tools.ietf.org/html/draft-yasskin-dispatch-web-packaging-00

### Other information
It will allow people to bundle together the resources that make up a website, so they can be shared offline, either with or without a proof that they came from the original website.

Repo: https://github.com/WICG/webpackage

overholt:
FWIW, today I was asked about this by Yoav and a few Googlers.

marcoscaceres:
Just noting the use of web packaging and AMP ([blog post](https://www.ampproject.org/latest/blog/improving-urls-for-amp-pages/)). We should evaluate what this means exactly and how web packaging addresses concerns around AMP.

jyasskin:
This isn't a complete answer to @hsivonen's concerns, but I figured I'd point you to the up-to-date proposals before you re-review anything:

https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html is additional HTTP-level metadata to receive exchanges without talking to the origin server. It's the piece that we think is most immediately useful for AMP. It proposes a way for browsers to opt into CDNs pushing cross-CDN content, although @yoavweiss suggested a second way that I haven't analyzed or written up yet. 

https://github.com/WICG/webpackage/pull/98 is the start of the MHTML replacement, and the PR lists a couple ways that it improves on MHTML.

https://tools.ietf.org/html/draft-yasskin-webpackage-use-cases-00 has a more complete list of use cases, and was published in August.

adamroach:
See also @ekr's comment at https://github.com/mozilla/standards-positions/pull/53#issuecomment-361710624

marcoscaceres:
Can someone kindly provide a link to the feedback @ekr is referenced in #53?

stpeter:
@marcoscaceres Seems to be https://lists.w3.org/Archives/Public/ietf-http-wg/2018JanMar/0071.html

ekr:
I'm finding myself a bit confused about which standards body this document seems to think it's targeted at. The link above is to a WICG repo but it's clearly an IETF draft.

jyasskin:
[...]

PR #53, opened 2018-01-30
Adding web packaging

adamroach:


ekr:
I've looked at this and I think there are major architectural concerns (see my recent message to HTTP). I haven't seen any compelling use cases that require this kind of major architectural change. Accordingly. I think we should mark this either "harmful" or "defer"

jyasskin:
Please don't call this "webpack". That's https://webpack.js.org/.

There's a collection of specifications that were called "web packaging", but they've been split up into at least "signed exchanges" at https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-02 and "bundles" or maybe "WebZip" which are draftily drafted at https://github.com/WICG/webpackage/pull/98.

It probably makes sense to add only "signed exchanges" right now, of course with whichever position @ekr and the rest of Mozilla negotiate.

dbaron:
We should keep the discussion about what our position is in #29.  That's the thing that will be referenced from the main table.

That said, I think it's worth answering @jyasskin's question about whether the item we're listing is the right one.  @adamroach - should we still go ahead with the pull request as-is, or does that make you want to list things differently?

(Also, maybe change the title of the PR if you can?)

adamroach:
>That said, I think it's worth answering @jyasskin's question about whether the item we're listing is the right one. @adamroach - should we still go ahead with the pull request as-is, or does that make you want to list things differently?

I'm not entirely certain what else I would point at in this regard. Are you proposing that we point to draft-yasskin-http-origin-signed-responses instead?

jyasskin:
[draft-yasskin-http-origin-signed-responses](https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses) is my suggestion.

dbaron:
@adamroach did you want to update this?

adamroach:
@dbaron -- I was planning on doing so after https://github.com/mozilla/standards-positions/issues/51 landed. I'd rather the disposition be from the new set rather than the old.

dbaron:
@adamroach ok, just (1 minute ago) landed that

activities.json
```diff
@@ -1,4 +1,15 @@
...
+    "mozBugUrl": null, 
+    "mozPosition": "harmful", 
+    "mozPositionDetail": "Mozilla has concerns about the shift in the web security model required for handling web-packaged infformation. Specifically, the ability for an origin to act on behalf of another without a client ever contacting the authoritative server is worrisome, as is the removal of a guarantee of confidentiality from the web security model (the host serving the webpackage has access to plaintext). We recognise that the use cases satisfied by web packaging are useful, and would be likely to support an approach that enabled such use cases so long as the foregoing concerns could be addressed.",
```
jyasskin: spelling: "infformation"

adamroach: Thanks.

dbaron:
In some cases I'll admit I've merged my own PRs after getting review from others.  But today I merged yours promptly, at least.

adamroach:
Thanks!

PR #56, opened 2018-02-06
Add new Firefox Logo + better align

MichaelKohler:


dbaron:
Thanks for the patch!

PR #57, opened 2018-02-17
Replace the `participating` position with `important`, `worth prototyping`, and `non-harmful`

### LINKED ISSUE #51: Set of possible positions seems too narrow

dbaron:
Fixes #51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ Implementation status (or even intention) isn't tracked here; see [Firefox Platf
 The currently possible positions are:
 
 - `under consideration` - Mozilla's position on this specification is being discussed.
-- `important` - This specification is important to Mozilla.
-- `worth prototyping` - Mozilla sees this specification as likely enough to be important that it is worth prototyping and iterating on.
-- `non-harmful` - Mozilla does not see this specification as harmful, but is not convinced that it is worth working on.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -24,7 +24,9 @@ Implementation status (or even intention) isn't tracked here; see [Firefox Platf
 The currently possible positions are:
 
 - `under consideration` - Mozilla's position on this specification is being discussed.
```
jyasskin: This might belong between `non-harmful` and `defer`, so that the positions are sorted by how favorable Mozilla is.

dbaron: I think I wanted to keep the current status of having `under consideration` at the start, since it's actually *not* a position.

jyasskin: Fair enough.

ekr:
LGTM

Issue #27, opened 2017-09-27
Web Share API


marcoscaceres:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Web Share API
* Specification or proposal URL: https://wicg.github.io/web-share/

### Other information
Currently behind a flag in Chrome as an "origin trial".

This specification defines an API for sharing text, links and other content to an arbitrary destination of the user's choice. The available share targets are not specified here; they are provided by the user agent. They could, for example, be apps, websites or contacts.

marcoscaceres:
Without claiming expertise in this area, I've reviewed the specification and raised a few small concerns with the design of the API (mostly just nits) - but overall, I think it's worth supporting from a standards perspective.   

The main concern I have is with the API appearing and disappearing depending on registered share targets. I've requested that the API just have `canShare()` method on it, which could be user configurable to just always return `false`. 
 
On the plus side, there are already Web Platform Tests for the API. And Google has run an origin trial for it. We can reach out to @mgiuca if we need additional information about the trail and site adoption. 

I'd also note that there is interest from the [Fennec side](https://bugzilla.mozilla.org/show_bug.cgi?id=1402369). Hi, @andreasbovens! 

We probably want into from @mixedpuppy, as he is really the authority on sharing at Mozilla. And @annevk has given this a lot of thought too, so would be great to get a 👍 from him.

mgiuca:
Some more info and links:

* Firefox bugs:
  * [1312422](https://bugzilla.mozilla.org/show_bug.cgi?id=1312422)
  * [1402369](https://bugzilla.mozilla.org/show_bug.cgi?id=1402369)
* This has since launched to stable on Chrome for Android. [[Intent to ship thread](https://groups.google.com/a/chromium.org/d/topic/blink-dev/nwmTLZKFFIA/discussion)]
* [Chrome platform status](https://www.chromestatus.com/features/5668769141620736)
* [Web Platform Tests](https://github.com/w3c/web-platform-tests/tree/master/web-share)
  * These are mostly manual tests (due to the nature of the UI). If implementing in Firefox, please ensure these tests pass manually. Let me know if you think there is a problem with the test.
* A major real-world test: [m.twitter.com](https://m.twitter.com) uses navigator.share. Just open any tweet, open its arrow menu, and click Share.

> The main concern I have is with the API appearing and disappearing depending on registered share targets. I've requested that the API just have `canShare()` method on it, which could be user configurable to just always return `false`.

[...]

Issue #24, opened 2017-09-27
RFP: Feature Policy


marcoscaceres:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Feature Policy
* Specification or proposal URL:  https://wicg.github.io/feature-policy/

### Other information
This is WICG proposal. 

A web platform API which gives a website the ability to allow and deny the use of browser features in its own frame, and in iframes that it embeds. Examples of features that could be controlled by feature policy include:

- getUserMedia (Camera, Speakers and Microphone)
- Fullscreen
- Geolocation
- MIDI
- Payments
- Synchronous XHR
- Synchronous scripts
- Vibrate

There is also a [detailed explainer](https://docs.google.com/document/d/1k0Ua-ZWlM_PsFCFdLMa8kaVTo32PeNZ4G7FFHqpFx4E/edit#heading=h.4yubgixv5l6b)

[Introduction to Feature Policy](https://developers.google.com/web/updates/2018/06/feature-policy)

annevk:
It would be good to figure this out, as proposals are made to integrate this into the various standards it affects. E.g., https://github.com/whatwg/fullscreen/issues/106.

I also believe that @kentuckyfriedtakahe and @blakewu likely want this for "autoplay" control of `<iframe>`s.

<hr>

Personally I think this approach seems reasonable. Offer control on a per-feature basis and leave grouping of features up to libraries and the future.

(It's a little unfortunate the permission proposal for not prompting on behalf of parties other than the one in the address bar got lost. Hopefully that makes a comeback some day.)

upsuper:
I read the draft briefly (without digging into every details), and it seems to me that it is a reasonable mechanism to control availability of features in documents.

It may need some investigation on how this would integrate / interact with other standards, what it means for future features, and whether its API is understandable and explorable for web developers.

martinthomson:
The general idea here is fine.  The implementation seems a little over-engineered.  For example, I don't see why this needs new HTTP header fields or a convoluted grammar.  I might be wrong, but I believe that script from a top-level browsing context always has a chance to run prior to things like `<iframe>`s being created, so it seems like a script-only solution is perfectly feasible.

dbaron:
> (It's a little unfortunate the permission proposal for not prompting on behalf of parties other than the one in the address bar got lost. Hopefully that makes a comeback some day.)

[...]

Issue #19, opened 2017-09-25
Permissions API


annevk:
* Specification Title: Permissions API
* Specification or proposal URL: https://w3c.github.io/permissions/

It seems that `request()` and `revoke()` were dropped so I suspect this is no longer controversial, but it came up in email with @marcoscaceres, @martinthomson, @jan-ivar, and @overholt.

The main question that came up in email is who gets to watch it.

dbaron:
Two policy/position issues I'd like to raise:

* I think it would be good to have something written down about our position on `request()` and `revoke()`, given that I've seen it come up a number of times that "Mozilla" has blocked those from being in the spec.  It would be useful to at least have the rationale for the opposition to those linked from this issue (since this issue will, in turn, be linked from the standards-positions table).

* I'd like to be clear that endorsement of the Permissions API doesn't imply endorsement of all of the things that it describes permissions for.  It's good that the spec is being thorough in that regard, but I don't want to include all of them in the process of evaluating it.

Also, one technical issue I'd like to bring up is whether we're OK with the `onchange` event handler mechanism on `PermissionStatus` and related lifetime issues.  It's not clear to me when the `PermissionStatus` and its event handler can/should be garbage-collected.  If the event handler has detectable side-effects (which seems likely to be the primary use of such an event handler, and therefore important to consider), it seems like garbage collecting the `PermissionStatus` and its event handler prior to the destruction of the page could lead to detectable differences, which seems to imply it should never be garbage-collected.  However, this is not how event handlers normally work; the presence of an event listener on an element normally doesn't keep the event alive.  But in this case it seems to require the awkward programming pattern of doing something explicitly to keep the `PermissionStatus` object alive so that the event handler will work; this pattern has been a footgun for authors in the past (at least in Gecko internal code, where I've seen it; I don't recall it being a thing on the Web).   Furthermore, if they are intended to be garbage-collected, are we OK with doing that via the cycle collector on workers these days?  And is the correct lifetime behavior here considered to be well-defined by the prose already in the specification, or does it need to be stated more explicitly? /cc @smaug----

It also seems like it might be less than ideal that the event handler doesn't get any information *about* the change, but instead has to run `query()` again.

There are various editorial things I could quibble with about the spec, such as:
* writing yet another definition of "powerful feature"
* inconsistency in the way the different permissions are presented (and the lack of precise use of the "boolean feature" definition)

I suppose I should file issues when I have time...

[...]

PR #66, opened 2018-03-05
Add the most active issues as "under consideration".

dbaron:
These issues seem worth trying to resolve first, and as part of being
more public about that discussion, I'm going to explicitly list them as
"under consideration" and reference their issues.

This will thus link to #58, #27, #24, and #19.

It also happens to include the first links to caniuse (the "ciuName"
field being non-null).

Note that activities.py normalized some whitespace in a prior entry
added in #53.  (I should figure out why it's always adding a space at
the end of the line, but that's a separate issue.  Until then, I'd
rather keep the file matching the way activities.py regenerates it
rather than having to hand-edit every time to avoid changing those
lines.)

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -4,12 +4,12 @@
     "description": "This document specifies how a server can send an HTTP request/ response pair, known as an exchange, with signatures that vouch for that exchange's authenticity. These signatures can be verified against an origin's certificate to establish that the exchange is authoritative for an origin even if it was transferred over a connection that isn't. The signatures can also be used in other ways described in the appendices. These signatures contain countermeasures against downgrade and protocol-confusion attacks.", 
     "mozBugUrl": null, 
     "mozPosition": "harmful", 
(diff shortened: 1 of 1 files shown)


martinthomson:
permissions-api mozPosition: remains complicated because that covers more than just `query()`

dbaron:
I'm actually hesitating about the bug URLs, because I'm a little hesitant to drive traffic to bugs.

dbaron:
I guess I should add the bugs at least, for the one that are in fact clearly metabugs for the entire spec.  I'm inclined to omit those that aren't.  I'll go through them later...

tantek:
<https://github.com/martinthomson> wrote:

> that covers more than just query()

In that case, shouldn’t we re-open <https://bugzilla.mozilla.org/show_bug.cgi?id=1105827>, and file new dependent bugs for what is not yet implemented / covered?

<https://github.com/dbaron> wrote:

> hesitating about the bug URLs, because I'm a little hesitant to drive traffic to bugs

Makes sense. Worth adding to this project readme/how-to officially? As in something like:

“The mozBugUrl field should only be filled in for "worth prototyping" or "important" standards, to avoid misdirecting higher-level discussion traffic to the bugs, which should be focused on implementation-specific discussions.”

(Originally published at: http://tantek.com/2018/064/t1/)

dbaron:
> In that case, shouldn’t we re-open <https://bugzilla.mozilla.org/show_bug.cgi?id=1105827>, and file new dependent bugs for what is not yet implemented / covered?

It's better not to repurpose bugs; the history becomes unreadable, and regression/dependency tracking becomes unusable.  If needed, a new bug should be filed, and if needed the old one could be retitled.

PR #67, opened 2018-03-06
Remove trailing spaces in activities.json.

dbaron:
I suspect editors like to strip out this trailing whitespace (e.g.,
from #53) which activities.py then readds when manipulating the JSON
(in #66).  We'll avoid the churn (and annoyance) by not having the
trailing whitespace at all.

This sets the separators as described in the documentation at
https://docs.python.org/2/library/json.html to avoid trailing whitespace
when using indent.

It also consolidates the JSON-writing configuration into a single place
by creating a single JSONEncoder object.


dbaron:
Er, I need to fix the newline at end of file.

PR #97, opened 2018-07-12
Add certificate compression for TLS

martinthomson:


ekr:
I agree with this proposed resolution.

PR #113, opened 2018-11-28
Improve cleaning of whitespace when importing specs.

dbaron:
This makes three changes to the whitespace cleanup:
* use `tag.strings` instead of `tag.stripped_strings` (which had the starting/ending whitespace of each tag removed), since use of the latter tended to remove whitespace around inlines
* instead of just replacing newlines with spaces, replace a newline *and any whitespace following it, including other newlines* with a space
* Remove starting and ending whitespace from the final result.

It also applies the whitespace cleaning to the title (in addition to the abstract).

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -289,7 +289,7 @@ class SpecParser(object):
         """
         Return a BeautifulSoup's tag contents as a string.
         """
-        return "".join(tag.stripped_strings).replace("\n", " ")
+        return re.sub("\n\s*", " ", "".join(tag.strings)).strip()
 
     @staticmethod
     def clean_url(url):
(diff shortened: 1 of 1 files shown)


dbaron:
I tested that this would have avoided all the manual cleanup that I had to do for the specs in #112.

activities.py
```diff
@@ -289,7 +289,7 @@ def clean_tag(tag):
         """
         Return a BeautifulSoup's tag contents as a string.
         """
-        return "".join(tag.stripped_strings).replace("\n", " ")
+        return re.sub("\n\s*", " ", tag.get_text()).strip()
```
martinthomson: Is it better to use `[\n\s]*`?  Or even `[\r\n\s]*`.  Or can we rely on the tags having whitespace collapsed and CRLF-normalized to `\n`?

dbaron: So I was intending to keep double-space after a period or colon when it was present -- although I wouldn't then create one at end-of-line, and this managed to do that, whereas replacing all whitespace with a single space wouldn't.  And as far as I can tell, (a) the only place there's odd whitespace we want to get rid of is at start-of-line, and (b) I haven't run into any `\r`s.  But I may run into odder specs in the future...

And note that `\n` is included within `\s`, which means `[ \t\n\r\f\v]` (or more when `UNICODE` is specified).  See https://docs.python.org/2/library/re.html .

PR #114, opened 2018-11-30
we are implementing CSS Grid 2

tantek:


tantek:
Issue filed: [https://github.com/mozilla/standards-positions/issues/125](https://github.com/mozilla/standards-positions/issues/125) and
updated pull request with mozPositionIssue and mozPositionDetail as requested.

(Originally published at: http://tantek.com/2018/353/t1/)

dbaron:
The grid part mostly looks good now -- although maybe it's worth working in something about "without removal of semantically important markup" or something like that into it.

However, your attempt to rebase against the intermediate changes didn't go so well, in that git still thinks you haven't rebased even though you have, so it's showing a very large and conflicting diff.

martinthomson:
Should we sort positions so that we don't get conflicts as a result of always adding to the end?

mnot:
See `entry_unique()`.

PR #133, opened 2019-01-24
Add cache digests as non-harmful

### LINKED ISSUE #131: PFP: Cache Digest

* Specification Title: Cache Digest for HTTP/2
* Specification or proposal URL: https://tools.ietf.org/html/draft-ietf-httpbis-cache-digest
 
### Other information

in httpwg at IETF, chair mnot asking implementation signal for cache digest now.

https://lists.w3.org/Archives/Public/ietf-http-wg/2019JanMar/0033.html

if mozilla says interested in or not, thats big material to consider this spec is dead or alive.

personally, I'm looking forward to use Cache Digest for Optimize http2 push.
if cache digest will dead, that means cache aware server push for browser dead too.
so I wanna hear good voice from firefox.

martinthomson:
As stated in the issue, this specification is highly experimental.
While it might be difficult to implement, the main problem is that it is
still unproven - like much of server push.

Closes #131.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -78,6 +78,17 @@
   },
   {
     "ciuName": null,
+    "description": "This specification defines a HTTP/2 frame type to allow clients to inform the server of their cache's contents. Servers can then use this to inform their choices of what to push to clients.",
+    "mozBugUrl": null,
+    "mozPosition": "non-harmful",
+    "mozPositionDetail": "This is experimental technology that might improve the use of server push by giving servers information about what is cached.  It is still unclear how much this might improve performance; more experimentation is likely necessary to prove this out.",
+    "mozPositionIssue": 131,
+    "org": "IETF",
+    "title": "Cache Digests for HTTP/2",
+    "url": "https://tools.ietf.org/html/draft-ietf-httpbis-cache-digest"
+  },
+  {
+    "ciuName": null,
     "description": "This document defines an imperative mechanism which allows web developers to instruct a user agent to clear a site's locally stored data related to a host.",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1268889",
     "mozPosition": "worth prototyping",


sebdeckers:
@martinthomson What would be acceptable as proof? It is trivial to demonstrate bandwidth and round-trip savings with POC implementations. Real-world evidence specific to HTTP is a chicken-egg problem.

Bloom/Cuckoo filters are widely used [in a variety of applications](https://en.wikipedia.org/wiki/Bloom_filter#Examples). AFAIK nothing suggests the concept does not apply to HTTP request/response round trips.

Consider that Apache mod_http2 ships with similar support built-in for a while now; not exactly the CD frame but a server side "push diary" to nullify over-push.

martinthomson:
Some research demonstrating efficacy in a variety of real-world cases would be good. That doesn't have to show improvements, but it would need to provide some insight into what factors need to be considered when deciding whether to deploy it, what conditions lead to gains and what situations are bad. With more information, we might be able to make an informed decision. All were have now is hearsay. Is there a paper describing the methodology and improvements from the trials that have been run?

sebdeckers:
## Research Paper

The 2014 Cuckoo Filter paper describes tuning to achieve fewer bits per entry for a desired false positive rate at an expected load rate and max number of entries.

- https://www.cs.cmu.edu/~dga/papers/cuckoo-conext2014.pdf
- https://github.com/efficient/cuckoofilter

## Demo

Cuckoo Filters as a server-side push diary. Tune the size of the Cuckoo filter to change its effectiveness.

https://gitlab.com/sebdeckers/demo-cache-digest-push-diary#readme

The demo uses the popular Font Awesome SVG icons library. These are ~1500 icons typically distributed in concatenated bundles in various formats: JS/CSS/SVG/fonts.

The code (see: `/index.js`) runs an HTTP/2 client and server in Node.js. The client receives the icon SVG files as PUSH_PROMISE/HEADER/DATA frames. Using a sufficiently sized "push diary", held in memory by the server, subsequent requests on the same client connection omit all subresources which are presumed to still be cached.

This would be improved upon by implementing Cache Digests in the client for several reasons:
- Re-use of digests across different connections. E.g. visit the webpage again tomorrow.
- Accurate reflection of purged/stale cached items on long-lived connections. Client can delete items from the CD and signal this to the server.
- 103 Early Hints: CDN edge servers or other middleboxes may use the digest to safely turn `Link: rel=preload` into PUSH_PROMISE without risking over-push, or strip known-fresh Link headers to reduce the 103 response headers size.

dbaron:
I'm going to merge this; the change seems technically fine.

Our practice is that the discussion of the *substance* of this position belongs in #131, not here.

PR #138, opened 2019-03-08
Provide guidance on updating positions

martinthomson:
This happens a bit and we should pave the path better.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,14 +71,13 @@ If successful, it will modify `activities.json` with the new specification. Chec
 ### Asking Mozilla to Update a Position
 
 Specifications might be updated after a position is decided. Especially if those
-updates are in response to feedback provided by Mozilla. If you think that
-changes to a proposal warrant an updated position from Mozilla, please request
-another review by commenting on the original issue for the specification. Please
-include a summary of the changes that you believe to be relevant.
+updates are in response to feedback provided by Mozilla.
 
-If you believe that you have new information that might cause Mozilla to
-reconsider a position, provide that information and ask a coordinator to reopen
-the issue.
+If you think that changes to a proposal warrant an updated position from
+Mozilla, please request another review by commenting on the original issue for
+the specification. Please include a summary of the changes that you believe to
+be relevant. A closed issue will be reopened if Mozilla agrees to make another
+assessment.
 
 ## Discussing Mozilla's Position on a Web Specification
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,10 +70,17 @@ If successful, it will modify `activities.json` with the new specification. Chec
 
 ### Asking Mozilla to Update a Position
 
-Specifications might updated after a position is decided. Updates might be in response to
-feedback provided by Mozilla. If you think that changes to a proposal warrant an updated
-position from Mozilla, please request another review by opening a new issue. Please include
-a summary of the changes that you believe to be relevant.
+Specifications might be updated after a position is decided. Especially if those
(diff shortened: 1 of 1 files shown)


dbaron:
So if you'd asked me to provide this advice, I'd have suggested reopening the existing issue, rather than filing a new issue.  (There's value in having the discussion in one place, particularly given that we link to the issue from the table.  Though maybe there's also value in separating the evaluations.)

I don't feel that strongly, though... what do you think?

martinthomson:
The only reason I didn't suggest that is that I wanted one piece of advice and the person asking us to reconsider might not be the one who asked in the first place.  I will amend the advice to allow reopening.

martinthomson:
On consideration, I think that reopening is ideal.  Our structure more or less depends on one issue per spec.  I've noted that coordinators should be the ones calling the shots on whether new information or changes warrant reassessment.

dbaron:
Two comments on the new version:

1. "Especially if those updates are in response to feedback provided by Mozilla." isn't a complete sentence and should somehow be restructured.

2. With the suggestion of not reopening, it might be worth suggesting @-mentioning somebody to get their attention -- but even then I'm not confident it will work.  I'd perhaps bias a little bit more towards reopening since I think without reopening there's a pretty significant chance we'll just miss the feedback entirely.

CONTRIBUTING.md
```diff
@@ -68,6 +68,19 @@ If successful, it will modify `activities.json` with the new specification. Chec
...
 
+### Asking Mozilla to Update a Position
+
+Specifications might be updated after a position is decided. Especially if those
+updates are in response to feedback provided by Mozilla. If you think that
+changes to a proposal warrant an updated position from Mozilla, please request
+another review by commenting on the original issue for the specification. Please
+include a summary of the changes that you believe to be relevant.
+
+When asking for an update you may reopen the issue. However, we ask that people
```
marcoscaceres: I’m unsure individuals from outside can reopen issues. Maybe say: “you can ask for an issue to be reopened”

annevk: They cannot, we should indeed rephrase that.

CONTRIBUTING.md
```diff
@@ -68,6 +68,17 @@ If successful, it will modify `activities.json` with the new specification. Chec
...
+
+Specifications might be updated after a position is decided. Especially if those
+updates are in response to feedback provided by Mozilla. If you think that
+changes to a proposal warrant an updated position from Mozilla, please request
+another review by commenting on the original issue for the specification. Please
+include a summary of the changes that you believe to be relevant.
+
+If you believe that you have new information that might cause Mozilla to
+reconsider a position, provide that information and ask a coordinator to reopen
+the issue.
```
dbaron: The second paragraph reads like it's just an alternative version of the first with slightly different advice.  Maybe merge the "ask a coordinator to reopen the issue" in to the end of the next-to-last sentence of the first paragraph and drop the rest?

("coordinator" also isn't all that well-defined...)

martinthomson:
Sorry for letting this drop.  I've made some more edits.

CONTRIBUTING.md
```diff
@@ -68,6 +68,16 @@ If successful, it will modify `activities.json` with the new specification. Chec
 }
 ```
 
+### Asking Mozilla to Update a Position
+
+Specifications might be updated after a position is decided. Especially if those
+updates are in response to feedback provided by Mozilla.
```
dbaron: This "especially" should be attached to something.  I don't think it's that specifications are updated especially when Mozilla provides feedback, though. :-)  I think it's something about updating the position.

PR #140, opened 2019-03-09
Add HTML Modules as worth prototyping.

### LINKED ISSUE #137: HTML Modules

* Specification Title: HTML Modules
* Specification or proposal URL:  https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/HTMLModules/explainer.md
* Mozillians who can provide input (optional): @annevk, @dbaron 

### Other information
Proposes "...an extension of the ES6 Script Modules system to include HTML Modules. These will allow web developers to package and access declarative content from script in a way that allows for good componentization and reusability, and integrates well into the existing ES6 Modules infrastructure."

Intent to implement in Blink:
https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/HTMLModules/i2i.md

Proposed changes to ES and HTML specs:
https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-module-spec-changes.md

Cc'ing @dandclark  @bocupp @samsebree @travisleithead from Microsoft, so they can track discussion and answer questions.

dbaron:
(What "org" to use here is a bit ambiguous.)

Fixes #137.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -110,6 +110,17 @@
     "url": "https://wicg.github.io/feature-policy"
   },
   {
+    "ciuName": null,
+    "description": "An extension of the ES6 Script Modules system to include HTML Modules. These will allow web developers to package and access declarative content from script in a way that allows for good componentization and reusability, and integrates well into the existing ES6 Modules infrastructure.",
+    "mozBugUrl": null,
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": null,
+    "mozPositionIssue": 137,
+    "org": "W3C",
+    "title": "HTML Modules",
+    "url": "https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-explainer.md"
+  },
+  {
     "ciuName": "client-hints-dpr-width-viewport",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -109,6 +109,17 @@
...
     "url": "https://wicg.github.io/feature-policy"
   },
+  {
+    "ciuName": null,
+    "description": "An extension of the ES6 Script Modules system to include HTML Modules. These will allow web developers to package and access declarative content from script in a way that allows for good componentization and reusability, and integrates well into the existing ES6 Modules infrastructure.",
+    "mozBugUrl": null,
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": null,
+    "mozPositionIssue": 137,
+    "org": "W3C",
```
annevk: I think the plan of record is a PR against the HTML Standard, so this should probably be WHATWG. We could also change it later though.

dbaron: I was going to... and then I realized that the system wasn't set up for "org": "WHATWG"... but we should probably fix that.

PR #162, opened 2019-05-29
Bind the pronoun "it" to "Mozilla"

spinda:


miketaylr:
(seems like I can't merge, but r+ all the same)

Issue #163, opened 2019-05-31
WebAuthn


mhalano:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Web Authentication: An API for accessing Public Key Credentials
* Specification or proposal URL: https://www.w3.org/TR/webauthn/
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): 

### Other information
This is the famous WebAuthn standard. Mozilla already have a working prototype on Firefox. This issue is just to formally register the support for standard.

dbaron:
Probably `important`.  cc @jcjones.

PR #181, opened 2019-07-05
Per Issue #163 - Indicate support for Web Authentication

jcjones:
Not sure how much to write in the detail section, but here's a first pass, per Issue #163.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -221,6 +221,17 @@
   },
   {
     "ciuName": null,
+    "description": "This specification defines an API enabling the creation and use of strong, attested, scoped, public key-based credentials by web applications, for the purpose of strongly authenticating users. Conceptually, one or more public key credentials, each scoped to a given WebAuthn Relying Party, are created by and bound to authenticators as requested by the web application. The user agent mediates access to authenticators and their public key credentials in order to preserve user privacy. Authenticators are responsible for ensuring that no operation is performed without user consent. Authenticators provide cryptographic proof of their properties to Relying Parties via attestation. This specification also describes the functional model for WebAuthn conformant authenticators, including their signature and attestation functionality.",
+    "mozBugUrl": null,
+    "mozPosition": "important",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -219,6 +219,17 @@
     "title": "Transport Layer Security (TLS) Certificate Compression",
     "url": "https://tools.ietf.org/html/draft-ietf-tls-certificate-compression"
   },
+  {
+    "ciuName": null,
```
dbaron: This should be "webauthn" to point to https://caniuse.com/#feat=webauthn

activities.json
```diff
@@ -219,6 +219,17 @@
...
+    "description": "This specification defines an API enabling the creation and use of strong, attested, scoped, public key-based credentials by web applications, for the purpose of strongly authenticating users. Conceptually, one or more public key credentials, each scoped to a given WebAuthn Relying Party, are created by and bound to authenticators as requested by the web application. The user agent mediates access to authenticators and their public key credentials in order to preserve user privacy. Authenticators are responsible for ensuring that no operation is performed without user consent. Authenticators provide cryptographic proof of their properties to Relying Parties via attestation. This specification also describes the functional model for WebAuthn conformant authenticators, including their signature and attestation functionality.",
+    "mozBugUrl": null,
```
dbaron: This should probably be "https://bugzilla.mozilla.org/show_bug.cgi?id=1294514"

activities.json
```diff
@@ -219,6 +219,17 @@
...
+    "mozPosition": "important",
+    "mozPositionDetail": "Public key cryptographic authentication is a major improvement in the fight against phishing, and we encourage all security-concious web applications to implement and innovate authentication flows around Web Authentication in the future.",
```
dbaron: The use of "innovate" here is a bit awkward since innovate is an intransitive verb and you're giving it a direct object...

martinthomson: Also typo conscious

jcjones: Aha, oops. Thanks!

PR #189, opened 2019-07-12
Add position on BigInt (non-harmful)

### LINKED ISSUE #65: RFP: BigInt

* Specification Title: BigInt
* Specification or proposal URL: http://tc39.github.io/proposal-bigint
* Caniuse.com URL (optional): None
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1366287

### Other information

BigInt provides a second numeric type in JavaScript: unlimited-size integers. In the Blink Intent to Ship thread for BigInt, @bzbarsky [noted](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/vIVL6eDQKf0/WEfVYnyRDQAJ) that, although @cxielarko has been developing a BigInt implementation for SpiderMonkey (as an employee of Igalia, as part of partnership Bloomberg), it has not yet been accepted by Mozilla engineers. It'd be great to have an official Mozilla position on whether BigInt should be added to the Web Platform.

cc @tschneidereit @syg @jorendorff @jswalden

marcoscaceres:
Closes #65

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -14,7 +14,7 @@
     "ciuName": null,
     "description": "This proposal adds arbitrary-precision integers to ECMAScript.",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1366287",
-    "mozPosition": "important",
+    "mozPosition": "worth prototyping",
     "mozPositionDetail": "Shipping in Firefox.",
     "mozPositionIssue": 65,
     "org": "Ecma",
--- a/activities.json
+++ b/activities.json
@@ -14,7 +14,7 @@
     "ciuName": null,
     "description": "This proposal adds arbitrary-precision integers to ECMAScript.",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1366287",
-    "mozPosition": "not harmful",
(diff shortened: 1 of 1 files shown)


dbaron:
I'm not sure `important` is justified by the thread in 65 -- and I don't think shipping the feature should imply `important`.  What position would we take if we hadn't yet shipped it?

martinthomson:
@dbaron is right here.  Yes, we're shipping it, but that is a decision that is influenced by factors other than the inherent value of the spec for the web, like the contract we enter into as part of TC39 participation.  A weaker position would do fine like `not harmful` might suffice.

`important` generally means that we think that the feature is an integral part of our overall vision.

activities.json
```diff
@@ -10,6 +10,17 @@
     "title": "Badging API",
     "url": "https://wicg.github.io/badging"
   },
+  {
+    "ciuName": null,
+    "description": "This proposal adds arbitrary-precision integers to ECMAScript.",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1366287",
+    "mozPosition": "not harmful",
```
```suggestion
    "mozPosition": "non-harmful",
```

PR #220, opened 2019-11-16
Make the JS icon show up for Ecma standards.

dbaron:
We had the JS icon in `asset/` but it wasn't showing up after we added the first Ecma position in #189.  This makes it show up.

### DIFF
--- a/index.html
+++ b/index.html
@@ -133,6 +133,9 @@
                     if (data === "IETF") {
                       extension = ".gif"
                     }
+                    if (data === "Ecma") {
+                      data = "JS"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -133,6 +133,9 @@ <h3>legend</h3>
                     if (data === "IETF") {
                       extension = ".gif"
                     }
+                    if (data === "Ecma") {
```
nschonni: Minor since it can't be both values
```suggestion
                    else if (data === "Ecma") {
```

Issue #20, opened 2017-09-26
Trusted Types


annevk:
* Specification Title: Trusted Types for DOM Manipulation
* Specification or proposal URL: https://github.com/WICG/trusted-types/blob/master/README.md

It seems useful to figure out something here early given its impact across a wide range of standards.

dbaron:
also cc @dveditz @jonathanKingston @mozfreddyb

jonathanKingston:
I have read this a few times already, besides that there might be some missing sinks (I haven't checked) it seems like a worthwhile addition to auditing.

I specifically like the CSP header to turn off all sinks, it might need to be more granular perhaps and could be in a FeaturePolicy perhaps?

We should also update https://github.com/mozilla/eslint-plugin-no-unsanitized to include these types.

My only questions/concerns are:
- Naming, does `TrustedHTML` imply that it will never permit XSS to developers?
- Does this have any impact to the XSS filter in the browser? Perhaps it could be disabled when this type is used?

mozfreddyb:
I find this an interesting approach, but I'm not sure a type system is solution the web has been looking for.

FWIW, I'd be more inclined to widely discuss the approach we have recently taken in Firefox's chrome privilege code: Automatically sanitize within the APIs that parse strings into HTML (e.g., innerHTML).
One could also debate exposing a sanitizer API to the DOM.

ekr:
@ckerschb

ckerschb:
After so many years XSS is still the most prevalent type of attacks on web applications and I agree that some kind of type system for manipulating the DOM could improve the situation. As mentioned in the explainer, security reviews of large applications have always been a pain point and I can see the benefit that security reviewers could focus their efforts around the creation of trusted types. Additionally we could enforce a variety of policies on the various types (maybe even within the DOM Bindings) which sounds tempting to me.

What I am mostly worried about is the policy creation. In the current form of the proposal a careless developer would most likely just register a no-op policy and it would be up to the security team to call that out - so this part would only be a minor improvement to the current string based APIs. Additionally I don’t think it should even be the developer to register the policy, wouldn’t it be better to separate the trusted type creation from the policy creation? Maybe we should try to think of some other policy creation and delivery mechanism. In turn, this shift of responsibility would remove complexity from the type creation system and the burden for developers to come up with a sophisticated policy. But maybe that would also require to build in some string sanitizers into the platform, which I personally would like to see anyway.

martinthomson:
[...]

PR #221, opened 2019-11-16
Add Trusted Types (#20) as 'defer'.

dbaron:


mikewest:
After the conversations we've had recently, I thought we'd ended up with a reasonable direction for the mechanism that y'all could, at least in theory, get on board with. I was hopeful that Mozilla would come down on "worth prototyping" or at least "non-harmful". Though I'm happy to see you haven't chosen "harmful", it surprises me a bit that y'all would end up punting the question.

+@annevk, +@jonathanKingston

dbaron:
Please have the discussion about what the position should be in #20 rather than here.

(It sounds like that discussion hasn't been reflected over there.)

mikewest:
@annevk, @jonathanKingston: Perhaps one of you could update the thread in #20 with your takeaways from our most recent discussion?

annevk:
Sorry about that, https://github.com/mozilla/standards-positions/issues/20#issuecomment-555016682.

dbaron:
I'll make a new PR for whatever we decide there.

PR #223, opened 2019-11-19
Add Fetch Metadata Request Headers position

### LINKED ISSUE #88: Fetch Metadata Request Headers
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1508292.

See https://github.com/whatwg/fetch/issues/700. Also mentioned in #80. Concerns mentioned in that issue:

> My concern with that header is that it makes referrer leaking worse, unless we restrict it to revealing one of same-origin, same-site, or cross-site.

Another concern is that it's yet another request header. It could perhaps be made opt-in, but that requires new infrastructure that hasn't materialized yet. As such I don't think this is something we can or want to do soon.

annevk:
Closes #88.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -144,6 +144,17 @@
   },
   {
     "ciuName": null,
+    "description": "This document defines a set of Fetch metadata request headers that aim to provide servers with enough information to make a priori decisions about whether or not to service a request based on the way it was made, and the context in which it will be used.",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1508292",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "This gives servers useful context about requests that can be used to mitigate various security issues. There are some thorny issues around embed/object that made agreeing on a design more complicated than expected.",
+    "mozPositionIssue": 88,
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -142,6 +142,17 @@
     "title": "Feature Policy",
     "url": "https://wicg.github.io/feature-policy"
   },
+  {
+    "ciuName": null,
+    "description": "This document defines a set of Fetch metadata request headers that aim to provide servers with enough information to make a priori decisions about whether or not to service a request based on the way it was made, and the context in which it will be used.",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1508292",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "This gives servers useful context about requests that can be used to mitigate various security issues. There are some thorny issues around embed/object that made agreeing on a design more complicated than expected.",
```
martinthomson: I would phrase this as "There are aspects of the design that we feel are quite complicated, [...]" and "but these are unavoidable" or "and we would like to see more work to simplify over time"

annevk: I reworded it a bit, more reasonable now?

PR #226, opened 2019-11-21
Update the Feature Policy position

annevk:
As well as an editorial change.

Given that Feature Policy overall is still in flux I suspect we want to keep #24 open though.


annevk:
Flipping the defaults is no longer the plan for Feature Policy. At least the way we want to use this is for enabling things we restrict in third-party contexts by default. If we already allow them we wouldn't add a corresponding feature policy identifier.

PR #227, opened 2019-11-21
Add position on Web Components

### LINKED ISSUE #60: Web Components
Raising this per suggestion from @dbaron in #59. I'm not really sure how to split this up since as per https://github.com/w3c/webcomponents/blob/gh-pages/README.md the situation is complicated.

* Shadow DOM: we're implementing "v1" and this is now part of the DOM and HTML standards. (Some parts ended up in CSS, UI Events, Touch Events.)
* Custom elements: we're implementing "v1" and this is now part of the DOM and HTML standards.
* HTML Templates: shipped and now part of the HTML standard.
* CSS changes: I'm not sure what the status is here.
* HTML Imports: this is now abandoned I think. We've opposed this in the past: https://hacks.mozilla.org/2014/12/mozilla-and-web-components/.

On top of that there's various new proposals which @smaug---- and I should know more about next week (there's a meeting). My thinking is that we'll open issues for those proposals, but that for the above (apart from the CSS bits) we don't really need anything.

annevk:
This covers the four initial pieces of Web Components, as discussed in #60. Some historical discussion can be found at https://hacks.mozilla.org/2014/12/mozilla-and-web-components/ and https://hacks.mozilla.org/2015/06/the-state-of-web-components/.

Closes #60.


annevk:
"worth prototyping" in a way is reasonable as all of these do still have open bugs (in impl and spec), but I could see adding "worth having" or "okay" or some such. "acceptable" sounds a little mean to me, as if we'd rather not have it, which I don't think is the intent? (I'm happy to add a PR for adding such a state if you think that's worthwhile and we can bikeshed there with everyone else.)

activities.json
```diff
@@ -142,6 +153,17 @@
     "title": "Feature Policy",
     "url": "https://wicg.github.io/feature-policy"
   },
+  {
+    "ciuName": "imports",
+    "description": "HTML Imports are a way to include and reuse HTML documents in other HTML documents.",
+    "mozBugUrl": null,
+    "mozPosition": "harmful",
+    "mozPositionDetail": "Mozilla anticipated that JavaScript modules would change the landscape here and rather invests in evolving that, e.g., through HTML Modules.",
```
dbaron: I think "rather invests" -> "would rather invest" or something like that.

activities.json
```diff
@@ -142,6 +153,17 @@
     "title": "Feature Policy",
     "url": "https://wicg.github.io/feature-policy"
   },
+  {
+    "ciuName": "imports",
+    "description": "HTML Imports are a way to include and reuse HTML documents in other HTML documents.",
+    "mozBugUrl": null,
+    "mozPosition": "harmful",
```
dbaron: For a "harmful" position I think it's also probably worth saying a little bit more than this about why the alternative solution is better.

annevk:
Was out for a bit, feedback should be addressed now.

PR #246, opened 2020-01-14
Add IDs to rows, add UI to expose them, and :target styles for the linked-to row.

### LINKED ISSUE #245: Make the dashboard linkable?
It would be nice if I could link to the Mozilla position on origin policy. Currently I can link to https://github.com/w3ctag/design-reviews/issues/127. But I cannot link to the part of https://mozilla.github.io/standards-positions/ which states the final position.

dbaron:
This adds IDs to rows, add UI to expose them, and :target styles for the linked-to row.

Fixes #245.

This is mostly done, but not quite ready yet.  I need to look into browser bugs with both anchor-scrolling and `:target` when the content is generated dynamically -- I think it's probably stuff that ought to work but doesn't.  But maybe I'm doing something wrong.

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -146,6 +146,11 @@ class ActivitiesJson(object):
                 errors.append("Entry %i is not a dictionary." % i)
             title = entry.get("title", "entry %i" % i)
             errors = errors + self.validate_entry(entry, title)
+            # This is *outside* validate_entry so that "add" can add an
+            # entry with an empty ID (which must then be filled in), but
+            # it will cause a validation error for other operations.
+            if entry.get("id", "") == "":
+                errors.append("{} includes has empty id".format(title))
         return errors
 
     def validate_entry(self, entry, title=None):
--- a/activities.py
+++ b/activities.py
@@ -57,7 +57,7 @@ class ActivitiesJson(object):
     """
 
     expected_entry_items = [  # (name, required?, type)
-        ("cui_name", False, StringType),
+        ("id", True, StringType),
         ("title", True, StringType),
         ("description", True, StringType),
         ("ciuName", False, StringType),
@@ -121,10 +121,14 @@ class ActivitiesJson(object):
             e["title"].lower().strip() for e in self.data
         ]:
             raise ValueError(
-                ["%s already contains %s" % (self.filename, entry["title"])]
+                ["%s already contains title %s" % (self.filename, entry["title"])]
+            )
+        if entry["id"] in [e["id"] for e in self.data]:
+            raise ValueError(
+                ["%s already contains id %s" % (self.filename, entry["id"])]
             )
         if entry["url"] in [e["url"] for e in self.data]:
-            raise ValueError(["%s already contains %s" % (self.filename, entry["url"])])
+            raise ValueError(["%s already contains url %s" % (self.filename, entry["url"])])
 
     def validate(self):
         """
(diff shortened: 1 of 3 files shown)


annevk:
I think Adam is right that we need to put the IDs in `activities.json` directly as everything else can change and is not necessarily unique. Making the ID automatically generated when you add a new entry seems good though. (And we should probably have a uniqueness check in the validator.)

dbaron:
Yeah, that's probably the right thing.  Now that I think I've actually gotten everything working, I'll redo this PR that way (and probably also split into better commits, so that it can be "rebase and merge"d in a clean way.

dbaron:
OK, I think this is ready for review now.

I didn't implement auto-creation of IDs in `activities.py`, though.  `./activities.py add` just creates an empty-string ID.  (Validation complains about duplicate IDs, though, so if somebody actually *commits* an empty string ID, then `activities.py add` will start failing because adding the new entry will fail to validate!)  I think this is OK for now, though it could be improved on later.

dbaron:
And **please note that I want to merge this with rebase and merge, not squash and merge**.

dbaron:
The FIXME was really about the lack of a11y (esp. keyboard support) for opening the details thing given that it exists entirely in a `click` handler... but you're right that it would probably be better to conflate linked-to and details-open rather than having 2 separate states.

dbaron:
I should also fix my link workaround, since I don't want it to re-scroll when re-sorting the table, but I also don't see anything better that DataTables exposes.

dbaron:
On the other hand:
* I don't think I want to break having two sets of details open at once, and
* I probably shouldn't spend more time rewriting this whole things

so maybe the current approach is ok?

(I did just  fix my link workaround.)

activities.py
```diff
@@ -121,10 +121,14 @@ def entry_unique(self, spec_entry):
             e["title"].lower().strip() for e in self.data
         ]:
             raise ValueError(
-                ["%s already contains %s" % (self.filename, entry["title"])]
+                ["%s already contains title %s" % (self.filename, entry["title"])]
+            )
+        if entry["id"] in [e["id"] for e in self.data]:
+            raise ValueError(
+                ["%s already contains id %s" % (self.filename, entry["id"])]
             )
```
adamroach: Given the default value for `id` is `""`, I think it would be advisable to check for an empty ID here as well. While this code will prevent two such entries from validating, my reading is that it will allow _one_ to do so, which seems like it could become a recurring source of frustration.

index.html
```diff
@@ -218,7 +245,16 @@ <h3>legend</h3>
         }
 
         // Add event listener for opening and closing details
-        $('#mozPositions').find('tbody').on('click', 'td', function () {
+        $('#mozPositions').find('tbody').on('click', 'td', function (event) {
+          var a = $(event.target).closest('a');
+          if (a.length > 0 && !a[0].parentNode.classList.contains("details-control")) {
+            // Do nothing if the click was on a link, except for the
+            // details-control.
```
adamroach: Thanks for fixing this! This had been a papercut that I kept meaning to look into.

activities.py
dbaron: So the difficult thing about this is that while I could straightforwardly do this -- I'd then need to write code to make up an ID when adding the entry rather than adding an empty one, because the "add" code currently gives an error if the addition doesn't validate.  And if my code has edge cases where it doesn't work, people would have to fight with that code to get `./activities.py add` to work.

adamroach: Ugh. Okay, fair point.

dbaron: Actually, I guess I can just put the check *outside* `validate_entry`

PR #252, opened 2020-01-16
Remove indications that this is a "pilot".

### LINKED ISSUE #251: End pilot?
The front page still claims the "pilot" disclaimer.  Are we ready to consider ending that?

dbaron:
Fixes #251.


dbaron:
Yeah, I threw that in intentionally, while checking that there wasn't any wording in the pilot notice that I wanted to keep.  (I think the wording of the pilot notice actually came from the README.md, not the index.html, since it matched other text in the README.md)

PR #266, opened 2020-02-08
Defer Contact Picker API position

### LINKED ISSUE #153: Contact Picker API

* Specification Title: Contact Picker API
* Specification or proposal URL: [Chrome Status page](https://www.chromestatus.com/feature/6511327140904960), [Repository & Explainer](https://github.com/beverloo/contact-api)
* Mozillians who can provide input (optional): @annevk @tantek (probably should add others, please @-mention)

This is something Google is working on that it would be useful for us to provide input on.

tantek:
Closes #153.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -41,7 +41,7 @@
     "id": "contact-picker",
     "mozBugUrl": null,
     "mozPosition": "defer",
-    "mozPositionDetail": "This API innnovates in some ways beyond several previous Contacts APIs, though leaves their source unspecified, and uses different properties than HTML autofill field names.",
+    "mozPositionDetail": "This API innnovates in some ways beyond several previous Contacts APIs, though uses different properties than HTML autofill field names.",
     "mozPositionIssue": 153,
     "org": "WICG",
     "title": "Contact Picker API",
--- a/activities.json
+++ b/activities.json
@@ -37,6 +37,18 @@
   },
   {
     "ciuName": null,
+    "description": "This proposal adds an API for prompting and querying the user’s contacts for one or more items with a handful of contact properties.",
+    "id": "contact-picker",
+    "mozBugUrl": null,
+    "mozPosition": "defer",
+    "mozPositionDetail": "This API innnovates in some ways beyond several previous Contacts APIs, though leaves their source unspecified, and uses different properties than HTML autofill field names.",
+    "mozPositionIssue": 153,
+    "org": "WICG",
+    "title": "Contact Picker API",
+    "url": "https://wicg.github.io/contact-api/spec/"
+  },
+  {
+    "ciuName": null,
     "description": "This draft defines additions to CSS Grid, primarily for the subgrid feature.",
     "id": "subgrid",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1349043",


activities.json
```diff
@@ -35,6 +35,18 @@
     "title": "BigInt",
     "url": "https://tc39.github.io/proposal-bigint"
   },
+  {
+    "ciuName": null,
+    "description": "This proposal adds an API for prompting and querying the user’s contacts for one or more items with a handful of contact properties.",
+    "id": "contact-picker",
+    "mozBugUrl": null,
+    "mozPosition": "defer",
+    "mozPositionDetail": "This API innnovates in some ways beyond several previous Contacts APIs, though leaves their source unspecified, and uses different properties than HTML autofill field names.",
```
marcoscaceres: The current text states facts about the API, but don't really relate to the position. How about: 


```suggestion
    "mozPositionDetail": "Although we see this API as a non-harmful addition to the Web, Mozilla defers working directly on this API at this time. We encourage the WICG community to continue to iterate on a data model that is well-suited for representing contacts in a manner that is both privacy preserving and OS independent.",
```

adamroach: Two quick comments -- (1) My understanding is that "defer" indicates that we are deferring _stating a position_ pending additional data or other developments. Your proposal takes it to mean something subtly different (deferring standards work, which is related but not exactly the same thing). (2) While I understand where you're coming from with the phrasing "working directly on this API," it's easy to see how external parties my read "working" as "implementing" rather than "contributing to its specification." I believe phrasing like "...working directly on this API specification..." would make this less likely.

marcoscaceres: Good suggestion re: 2.. and also good points about 1. "Defer" is indeed, "Mozilla does not intend to look at this specification at all in the near future." I guess I got confused because we did look at it, but now we won't be investing any more time looking at it at all. I'm indeed biased as WICG Chair, as I do duck in an out of a lot of WICG specs as part of my day to day work.

dbaron: Yeah, I think Marcos's suggestion doesn't fit with a "defer" position.  I'd be OK with the part about moving it into the description, though.  That would leave the "detail" part *almost* blank -- though Tantek has a "leaving the source unspecified" that Marcos didn't copy, and I'm also not sure what that means.  @tantek should we drop or move that part as well?

marcoscaceres: > Yeah, I think Marcos's suggestion doesn't fit with a "defer" position. 

Re-checked my assumptions there. We should maybe clarify that a bit in the README where "defer" is defined... if I got it wrong, then likely others internally and externally folks are interpreting it incorrectly. 

There still needs to be something that says "we think this is in good hands, even though we don't have the resources to participate right now." Or is that not a thing? 

> Tantek has a "leaves their source unspecified" .. I'm also not sure what that means

I wasn't sure either... I was guessing that means the OS or app (e.g., Contact.app)? Revealing that would be weird tho. @tantek?

tantek:
I agree with [https://github.com/dbaron](https://github.com/dbaron) about Marcos’s suggestion doesn’t fit with a "defer" position.

I also strongly disagree with even hinting that we might think this API is "non-harmful" when that hasn’t even been mentioned in [https://github.com/mozilla/standards-positions/issues/153](https://github.com/mozilla/standards-positions/issues/153) — in fact quite the opposite, I noted explicitly there that "I’m leaning towards expecting an eventual 'harmful' evaluation."

So I’m not ok with moving it into the description, and that is unnecessary for a defer position. I’d like us to be able to actually defer rather than debate more details in the description.

Re: "leaves their source unspecified", I’m ok with dropping that from the description, it’s not really an actionable criticism outside of a thorough evaluation. I’ll update the PR accordingly.

(Originally published at: https://tantek.com/2020/042/t3/)

marcoscaceres:
>  I noted explicitly there that "I’m leaning towards expecting an eventual 'harmful' evaluation."

Why? It does a reasonable job at being privacy preserving - or did I miss something?

tantek:
&gt; Why? It does a reasonable job at being privacy preserving - or did I miss something?

Please ask follow-up questions of a potential eventual "harmful" evaluation like that on [https://github.com/mozilla/standards-positions/issues/153](https://github.com/mozilla/standards-positions/issues/153), rather than asking here on a pull request to "defer".

I’m also requesting "us to be able to actually defer rather than debate" a specific eventual evaluation.

(Originally published at: https://tantek.com/2020/042/t4/)

marcoscaceres:
Will follow up in the bug if need be. Thanks again @tantek for putting this together!

dbaron:
I merged this as-is.  On thinking further about `description` versus `mozPositionDetail` I realized we've generally used the `description` straight from the spec's own abstract when that text was appropriate (it's what `activities.py` does by default), so I think leaving the slight bit of editorializing in the detail seemed like the right thing.

PR #270, opened 2020-02-13
link template for discoverability

tantek:


nschonni:
Not sure this really helps, since the template is already used when creating a new issue https://github.com/mozilla/standards-positions/issues/new

dbaron:
If anything, I think it's better to link to https://github.com/mozilla/standards-positions/issues/new

dbaron:
I think it's still potentially confusing to link to the template since, at that URL, it's not in a form that somebody can do something useful with it.

PR #273, opened 2020-02-21
Make validate check that the entries are sorted correctly.

dbaron:
This was another oversight in #127.  I recently had to fix such an error in ad783e147f4643b4319172325220c85c0a1c3588 to avoid having extra diffs in my patches.

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -139,6 +139,7 @@ class ActivitiesJson(object):
         if not isinstance(self.data, list):
             return ["Top-level data structure is not a list."]
         errors = []
+        prevTitle = None
         i = 0
         for entry in self.data:
             i += 1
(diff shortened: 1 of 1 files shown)


activities.py
```diff
@@ -151,6 +152,12 @@ def validate(self):
             # it will cause a validation error for other operations.
             if entry.get("id", "") == "":
                 errors.append("{} includes has empty id".format(title))
+
+            # Check that the entries are sorted by title, as save writes them.
+            if i != 1:
+                if prevTitle > title:
```
martinthomson:
```python
if prevTitle is not None and prevTitle > title:
```
?  (Personal dislike of 1-based indexing there).

PR #274, opened 2020-02-24
Add WICG File Handling (defer)

### LINKED ISSUE #158: File Handling

* Specification Title: File Handling
* Specification or proposal URL: [Explainer](https://github.com/WICG/file-handling/blob/master/explainer.md), [Repository](https://github.com/WICG/file-handling)
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): @annevk @bzbarsky @hsivonen @dbaron 

### Other information

This proposal was just submitted for TAG review, and it seems at a high level like something I'd be interested in.  I haven't looked at the details yet.  It's still explainer-only; no spec yet.

annevk:
Closes #158.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -277,6 +277,18 @@
   },
   {
     "ciuName": null,
+    "description": "This proposal gives web applications a way to register their ability to handle (read, stream, edit) files with given MIME types and/or file extensions.",
+    "id": "wicg-file-handling",
+    "mozBugUrl": null,
+    "mozPosition": "defer",
+    "mozPositionDetail": "Not far enough along to properly evaluate.",
+    "mozPositionIssue": 158,
+    "org": "Proposal",
+    "title": "File Handling",
+    "url": "https://github.com/WICG/file-handling/blob/master/explainer.md"
+  },
+  {
+    "ciuName": null,
     "description": "A proposal for using URL fragments with spaces in them to select a bit of text to highlight and scroll to",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -227,6 +227,18 @@
...
   },
+  {
+    "ciuName": null,
+    "description": "This proposal gives web applications a way to register their ability to handle (read, stream, edit) files with given MIME types and/or file extensions.",
+    "id": "wicg-file-handling",
+    "mozBugUrl": null,
+    "mozPosition": "defer",
+    "mozPositionDetail": "Not far enough along to properly evaluate.",
+    "mozPositionIssue": 158,
+    "org": "Other",
```
```suggestion
    "org": "W3C",
```
Is this really what we want for WICG? Seems to give it more legitimacy than it really has.

jyasskin: Drive-by +1 to distinguish CG/BG documents from WG documents. They are still part of the W3C, so "Other" doesn't really fit either.

dbaron: For consistency with existing practice this particular commit should probably be `"W3C"`, but you're right that we should distinguish better.  I filed #278 on that, but if this lands first, we should probably do `"W3C"` here for consistency (and so it gets caught in the changes there).

annevk:
@dbaron I rebased this and changed the org to Proposal. Okay to merge this or do you think we should investigate more?

PR #277, opened 2020-02-24
Add domenic's Origin Isolation (worth prototyping)

### LINKED ISSUE #244: Origin isolation

* Specification Title: Origin isolation
* Specification or proposal URL: https://github.com/domenic/origin-isolation (this will eventually be merged into the [origin policy spec](https://wicg.github.io/origin-policy/) (#160) and the HTML Standard)
* Mozillians who can provide input: @annevk has been quite helpful so far.

### Other information

This proposal is a bit subtle. The normative impacts are relatively small: a site can opt out same-site cross-origin DOM object access via `document.domain`, and same-site cross-origin `postMessage()`ing of `SharedArrayBuffer`s. But the intent is that, while doing so, the site can provide hints on why it's giving away these capabilities, the browser can use those hints to guide its process allocation strategies.

As for each of those hints: Chrome has partners interested in `prefer_isolated_event_loop` and `for_memory_measurement`. I think we all recognize that side-channel protection is good, so there's also `for_side_channel_protection`.

I also included `prefer_isolated_memory`, since I thought Mozilla might appreciate a standardized version of their [`Large-Allocation`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Large-Allocation) header; thoughts welcome there. I imagine it doesn't line up exactly, but it seems to cover similar use cases...

We also have partners that will not be able to deploy COOP + COEP + CORB-across-all-their-dependencies, but are still interested in isolated event loops. So we think this is additive to the idea of [automatic origin isolation via COOP + COEP](https://github.com/domenic/origin-isolation#automatic-origin-isolation-via-coop--coep).

Although we're reasonably happy with where the explainer/spec draft has ended up right now, it's hard to say for sure what will survive real-world testing on partner sites (which we hope to do around the end of Q1). As such, any early design feedback would be much appreciated.

annevk:
Closes #244.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -325,6 +325,18 @@
   },
   {
     "ciuName": null,
+    "description": "Origin isolation refers to segregating cross-origin documents into separate agent clusters.",
+    "id": "domenic-origin-isolation",
+    "mozBugUrl": null,
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "Letting developers opt out of document.domain and reduce the potential size of their agent cluster allows user agents to balance security, performance, and resource management.",
+    "mozPositionIssue": 244,
+    "org": "Other",
+    "title": "Origin Isolation",
+    "url": "https://github.com/domenic/origin-isolation"
+  },
+  {
+    "ciuName": null,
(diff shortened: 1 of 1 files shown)


domenic:
Thank you! Sorry for not moving this to WICG; I'll work on the process for that soon.

Issue #278, opened 2020-02-24
Distinguish community group (CG) documents (including WICG) from rest of W3C


dbaron:
In the discussion in #274 it was suggested that we should probably distinguish documents in W3C Community Groups (and Business Groups), including WICG, from documents in W3C Working Groups.

I think this is a good idea and we should do it.

adamroach:
If we want a WICG-specific logo, there's an SVG version at https://wicg.io/images/logo.svg ![WICG logo](https://wicg.io/images/logo.svg)

adamroach:
And the general CGBG logo is at https://www.w3.org/community/src/templates/wordpress/StoryTeller/img/svg/cgbg-logo.svg ![CGBG Logo](https://www.w3.org/community/src/templates/wordpress/StoryTeller/img/svg/cgbg-logo.svg)

mnot:
+1 to CGBG rather than WICG - using the latter would legitimise it unreasonably.

dbaron:
The icon I had in my PR in #279 (prior to @adamroach's comment) was the latter CG/BG icon.

dbaron:
Based on followup discussion, I think we instead (of what I proposed initially here and started changes for in #279) want a category for anything that is not yet on a multi-vendor standards track. This can include W3C Community Groups, IETF individual submissions, or similar, or things in personal repositories.

I'm struggling a little bit with a clear one-word name for this, though.  (But I recall thinking of one when I wasn't in front of my computer, and then forgetting it...)

mnot:
What implication does that have for an Internet-Draft that's not yet adopted by a WG, or is pre-charter work?

cwilso:
If this is "not yet in a WG", that makes sense.  If it is some other definition, that definition should be really clear.

adamroach:
>I'm struggling a little bit with a clear one-word name for this, though. (But I recall thinking of one when I wasn't in front of my computer, and then forgetting it...)

Brainstorming:
- Unadopted
- Pre-process
- Early proposal
- Sketch
- Preliminary

martinthomson:
> What implication does that have for an Internet-Draft that's not yet adopted by a WG, or is pre-charter work?

For Internet-Drafts, anything without draft-ietf- fits in this new category of "stuff with no formal standing".

As for the word to use, I would just go with "proposal".

PR #279, opened 2020-02-24
Distinguish W3C Community Group documents from Working Group documents.

dbaron:
This is a draft to do #278.  I still need to test a bunch of things and double-check the licensing on the image, but wanted to post it to avoid duplication of work.


dbaron:
I actually intend to abandon this in favor of an alternative (see further discussion in #278).

PR #282, opened 2020-02-27
Reflect origin isolation's move to WICG

domenic:
I thought this might be appreciated :)


dbaron:
Thanks for the patch.

dbaron:
... and for moving it to WICG.

PR #284, opened 2020-03-04
Distinguish W3C Community Group documents (as proposals) from Working Group documents

### LINKED ISSUE #278: Distinguish community group (CG) documents (including WICG) from rest of W3C

dbaron:
Fixes #278.

I'm going to need someone else to go through the IETF documents, since I'm not sure which of the inconsistent indicators to use to determine which things are in a working group versus an individual submitter document.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -31,7 +31,7 @@
     "mozPosition": "worth prototyping",
     "mozPositionDetail": null,
     "mozPositionIssue": 108,
-    "org": "W3C",
+    "org": "Proposal",
     "title": "Badging API",
     "url": "https://wicg.github.io/badging"
   },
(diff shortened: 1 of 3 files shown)


dbaron:
@martinthomson So something like https://tools.ietf.org/html/draft-privacy-pass that says "Network Working Group" at the very top of the document is really an individual submitter document?

martinthomson:
That is correct.  That document is an individual submission.  The usual pattern is `draft-<surname>-<working group>-stuff`, but that one omits the surname piece (and the working group, but there isn't one yet).  A "<surname>" of "ietf" indicates that the document has been formally adopted.

dbaron:
What about something like https://tools.ietf.org/html/rfc5861 which is a Independent Submission RFC?

PR #285, opened 2020-03-04
Update a few URLs for IETF documents to current drafts.

dbaron:
I believe (although I'm not 100% sure that this is right) that the tools.ietf.org UI is telling me that these are newer versions of these documents.


dbaron:
And I should note I was starting to do this in order to do an IETF version of #284, but I wasn't sure what the right distinction to draw was, so I just did this.

PR #286, opened 2020-03-05
Convert IETF individual submissions from "IETF" to "Proposal", as for W3C Community Groups.

dbaron:
This is done based on the discussion (and help from @martinthomson) in #284, and fixes the IETF-part of #278.

This needs to land after #284.


dbaron:
Also note that I'm not sure that handling of RFCs is fully correct, though I'm pretty sure at least some of them should be `IETF`, and I suspect it's likely a reasonable default.  https://github.com/mozilla/standards-positions/pull/284#issuecomment-595064272 has the ones I'm not sure about.

PR #291, opened 2020-03-06
Add Trusted Types (non-harmful)

### LINKED ISSUE #20: Trusted Types

annevk:
Closes #20.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -553,6 +553,18 @@
   },
   {
     "ciuName": null,
+    "description": "An API that allows applications to lock down powerful APIs to only accept non-spoofable, typed values in place of strings to prevent vulnerabilities caused by using these APIs with attacker-controlled inputs.",
+    "id": "trusted-types",
+    "mozBugUrl": null,
+    "mozPosition": "non-harmful",
+    "mozPositionDetail": "The API does not seem suitable for the long tail of sites.",
+    "mozPositionIssue": 20,
+    "org": "W3C",
+    "title": "Trusted Types",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -551,6 +551,18 @@
     "title": "Trust Token API",
     "url": "https://github.com/WICG/trust-token-api/blob/master/README.md"
   },
+  {
+    "ciuName": null,
+    "description": "An API that allows applications to lock down powerful APIs to only accept non-spoofable, typed values in place of strings to prevent vulnerabilities caused by using these APIs with attacker-controlled inputs.",
+    "id": "trusted-types",
+    "mozBugUrl": null,
+    "mozPosition": "non-harmful",
+    "mozPositionDetail": "The API does not seem suitable for the long tail of sites.",
```
martinthomson: I might be more forthcoming with this: The API could be used to harden sites against certain cross-site scripting issues, but it is sufficently complex that we are concerned that it will not be suitable for many sites.

```suggestion
    "mozPositionDetail": "The API could be used to harden sites against certain cross-site scripting issues, but it is sufficiently complex that we are concerned that it will not be suitable for many sites.",
```

franktopel:
I'm sorry to jump in here, what does Mozilla considering the feature `non-harmful` actually mean for the implementation prospects of `TrustedTypes`?

nschonni:
https://github.com/mozilla/standards-positions#possible-specification-positions

PR #292, opened 2020-03-07
Center the organization logos.

dbaron:
This is a followup to #284 and #286, which made wider use of a narrower organization icon for proposals.

I think the resulting table looks better when that column is centered.


dbaron:
And for the record, the documentation that led me to this was https://datatables.net/manual/styling/classes

PR #293, opened 2020-03-10
Web Share API is now in the Web Applications WG.

dbaron:


marcoscaceres:
See my comment about "worth prototyping" in the related bug. But keeping it "under consideration" is ok too.

PR #295, opened 2020-03-20
Make the 'sort' command work again.

dbaron:
I broke this in #273.


martinthomson:
Oops.  I guess it's hard to notice this sort of thing when all the paved paths lead to not hitting the problem.

PR #309, opened 2020-04-10
Add Raw Clipboard Access (harmful).

### LINKED ISSUE #206: Raw Clipboard Access API

* Specification Title: Async Clipboard: Raw Clipboard Access
* Specification or proposal URL: https://github.com/dway123/raw-clipboard-access/blob/master/explainer.md
* TAG design review URL: https://github.com/w3ctag/design-reviews/issues/406
* Mozillians who can provide input (optional): @smaug----

### Other information

This was a topic during [the Editing TF meeting](https://lists.w3.org/Archives/Public/public-editing-tf/2019Oct/0004.html) at TPAC this year.

Related to, but distinct from #89

dbaron:
Closes #206.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -457,6 +457,18 @@
   },
   {
     "ciuName": null,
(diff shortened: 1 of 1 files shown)


rolivo:
Firefox is a bad thing for this world since they do not manage to implement APIs at the same rate as chrome, they are dedicated only to repressing the use of these using security and privacy as a shield.
Firefox are nothing but a bunch of desperate bad people who can't compete with chrome.

PR #310, opened 2020-04-10
Add Keyboard Map API (harmful).

### LINKED ISSUE #300: Position on Keyboard Map API

* Specification Title: Keyboard Map API
* Specification or proposal URL: https://wicg.github.io/keyboard-map/
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1469017
* Mozillians who can provide input (optional): @masayuki-nakano @smaug----
### Other information
This is not possible with existing web platform APIs because the value that should be shown to the user depends on the keyboard layouts that the user has installed and active.

Explainer: https://github.com/WICG/keyboard-map/blob/master/explainer.md

dbaron:
Closes #300.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -360,6 +360,18 @@
     "url": "https://tools.ietf.org/html/draft-ietf-httpbis-client-hints"
   },
   {
+    "ciuName": null,
+    "description": "This specification defines an API that allows websites to convert from a given code value to a valid key value that can be shown to the user to identify the given key. The conversion from code to key is based on the user\u2019s currently selected keyboard layout. It is intended to be used by web applications that want to treat the keyboard as a set of buttons and need to describe those buttons to the user.",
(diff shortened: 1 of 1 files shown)


dbaron:
Oh, hey, I finally figured out what I needed to do configure to let me request reviews from @bholley, so I'll do that for this one.  (It was what I was originally trying to do here but couldn't figure out how!)  (I needed to make him a "Member" of the repository, which is something I do have the power to do.)

rolivo:
Firefox is a bad thing for this world since they do not manage to implement APIs at the same rate as chrome, they are dedicated only to repressing the use of these using security and privacy as a shield.
Firefox are nothing but a bunch of desperate bad people who can't compete with chrome.

PR #316, opened 2020-04-15
Add netinfo API (harmful)

### LINKED ISSUE #117: Network Information API

* Specification Title: Network Information API
* Specification or proposal URL: https://wicg.github.io/netinfo/
* Caniuse.com URL (optional): https://caniuse.com/#feat=netinfo
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): @martinthomson, @marcoscaceres, @tomrittervg

### Other information

This API ships in Chrome.  Firefox had an implementation of the API, which has subsequently been removed.  [This dev.platform thread from 2016](https://groups.google.com/d/msg/mozilla.dev.platform/lCZmhCDGHPY/hq3CgPkFDwAJ) covers this.  Some of that discussion is might be stale as the spec has changed a little in the meantime.

Of note, since those discussions, the `Save-Data` feature was moved to this specification.  This is a signal to a site that the client prefers a trade-off in favour of network efficiency when there are choices to be made between the number of bytes transferred and other things (like CPU overheads, quality, etc...)

martinthomson:
Closes #117.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -426,7 +426,7 @@
     "mozBugUrl": null,
     "mozPosition": "harmful",
     "mozPositionDetail": "This API provides information about a client's network connection, which allows sites to obtain additional information about clients and their environment. It is better that sites use methods that dynamically adapt to available bandwidth, as that is more accurate and likely to be applicable in the moment.",
-    "mozPositionIssue": null,
+    "mozPositionIssue": 117,
     "org": "Proposal",
     "title": "Network Information API",
     "url": "https://wicg.github.io/netinfo"
--- a/activities.json
+++ b/activities.json
@@ -420,6 +420,18 @@
     "url": "https://wicg.github.io/native-file-system"
   },
   {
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -419,6 +419,18 @@
     "title": "Native File System",
     "url": "https://wicg.github.io/native-file-system"
   },
+  {
+    "ciuName": "netinfo",
+    "description": "The Network Information API enables web applications to access information about the network connection in use by the device.",
+    "id": "netinfo",
+    "mozBugUrl": null,
```
marcoscaceres, line 426: Maybe https://bugzilla.mozilla.org/show_bug.cgi?id=1057169 ?

marcoscaceres: we could "won't fix" that also.

martinthomson: Thanks for the link.  We probably should, but that's a separate decision.

annevk: I couldn't resist.

PR #317, opened 2020-04-15
Use Python 3

martinthomson:
This might be a little disruptive if people have been using python2, but it's probably worth making this official.

(For those on distros with the bare `python` linked to `python2`, this is a change, unless they took steps to only use `python3`, as I have done.)


mnot:
+1 - it is time.

PR #328, opened 2020-04-19
Open details for the target when the hash changes.

### LINKED ISSUE #323: Fragment links should expand the details
As a followup to #245, when I link to https://mozilla.github.io/standards-positions/#webusb, it would be nice if it would auto-expand the Description and Mozilla's Position `<tr>` instead of making the person who gets the link figure out which piece of the row to click to get more details.

dbaron:
Fixes #323.

### DIFF
--- a/index.html
+++ b/index.html
@@ -258,7 +258,7 @@
         function open_row(row, tr) {
           var extra_data = format(row.data());
           if (extra_data !== "") {
-            row.child( format(row.data()) ).show();
+            row.child( extra_data ).show();
             tr.classList.add('shown');
           }
         }
--- a/index.html
+++ b/index.html
@@ -255,6 +255,19 @@
           }
         }
 
+        function open_row(row, tr) {
+          var extra_data = format(row.data());
+          if (extra_data !== "") {
+            row.child( format(row.data()) ).show();
+            tr.classList.add('shown');
+          }
+        }
+
+        function close_row(row, tr) {
+          row.child.hide();
+          tr.classList.remove('shown');
+        }
+
         // Add event listener for opening and closing details
         $('#mozPositions').find('tbody').on('click', 'td', function (event) {
           var a = $(event.target).closest('a');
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -266,20 +279,15 @@ <h3>legend</h3>
...
 
-          var tr = $(this).closest('tr');
+          var tr = $(this).closest('tr')[0];
           var row = ptable.row( tr );
           if ( row.child.isShown() ) {
             // This row is already open - close it
-            row.child.hide();
-            tr.removeClass('shown');
+            close_row(row, tr);
           }
```
martinthomson, line 287: Ugh, I just saw how bootstrap (?) is doing this details row.  That's just nasty.  I was going to say that you wanted a class on `<tr>` for "shown" items and you could just use `tr.classList.remove/add/toggle('shown')` to manage this.  But that isn't how this is built.

If you put the details into the DOM always, nested under identified rows, you might even use ':target' rules and avoid the need for most of the logic (only the open/close logic for clicks).  It might even be more accessible as a result (as the current method for accessing that information is not especially accessible.

index.html
```diff
@@ -255,6 +255,19 @@ <h3>legend</h3>
...
         }
 
+        function open_row(row, tr) {
+          var extra_data = format(row.data());
+          if (extra_data !== "") {
+            row.child( format(row.data()) ).show();
```
martinthomson, line 261:
```suggestion
          var extra_data = format(row.data());
          if (extra_data !== "") {
            row.child( extra_data ).show();
```

No point calling the same function twice, right?

index.html
```diff
@@ -255,6 +255,19 @@ <h3>legend</h3>
...
+            tr.classList.add('shown');
+          }
+        }
+
+        function close_row(row, tr) {
+          row.child.hide();
+          tr.classList.remove('shown');
```
martinthomson, line 268: As far as I can tell, the 'shown' class isn't really doing anything.

dbaron: It's used by the CSS to style the row after the target when the item is expanded (i.e., give it the yellow background).  (And that is still "needed" since it's possible to close an item while it remains the target.)

index.html
dbaron: It's not entirely clear to me what would be more accessible -- it may well be better as it is, at least assuming that the "Details" button (the (i) icon) is sufficiently accessible, since I'm not sure you'd want accessibility tools to have to read all the details in all cases -- although I suspect they actually wouldn't anyway since it would be `display:none`, and it wouldn't make a difference.  But I'm really not entirely sure.

That said, I was trying to just do this with minimal changes, without digging in too much.  (But now I dug in a bit more, and found the [documentation for `child`](https://datatables.net/reference/api/row().child()) and [for `show`](https://datatables.net/reference/api/row().child.show()) which are part of DataTables.)

So I'm somewhat hesitant to try to do this, given that it seems like the sort of thing that might lead to ripping out DataTables entirely and redoing this without it -- which while likely a reasonable design -- was more than I wanted to bite off today.

martinthomson: Yeah, I get that.  This is fine as is.  I just found this to be contrary to my experience building similar UI.  (I'm not defending that experience, BTW.)

PR #334, opened 2020-04-28
Add CORB (non-harmful)

### LINKED ISSUE #81: Fetch: CORB
CORB (cross-origin resource blocking) is a proposal by Google to disallow more responses to cross-origin "no-cors" requests: https://github.com/whatwg/fetch/issues/681.

Given Spectre and issues with "no-cors" in general this seems worth doing to me, but I'd like to double check.

If we want to mark these smaller proposals, I'd suggest `worth-prototyping`.

cc @ckerschb

annevk:
Note that I have added this as Proposal as it ought to be removed from Fetch due to its single implementer status.

Closes #81.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1,14 +1,14 @@
 [
   {
     "ciuName": null,
-    "description": "Adds imagesrcset and imagesizes attributes to <link> which correspond to the srcset and sizes attributes of <img> respectively, for the purposes of preloading.",
+    "description": "Adds imagesrcset and imagesizes attributes to &lt;link> which correspond to the srcset and sizes attributes of &lt;img> respectively, for the purposes of preloading.",
     "id": "image-preload",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1515760",
     "mozPosition": "worth prototyping",
-    "mozPositionDetail": "A relevant aspect of <link rel=preload> support.",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -263,6 +263,18 @@
     "title": "Credential Management Level 1",
     "url": "https://w3c.github.io/webappsec-credential-management/"
   },
+  {
+    "ciuName": null,
+    "description": "Blocklist certain opaque responses based on MIME type and return an 'emptied' response instead.",
+    "id": "corb",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1459357",
+    "mozPosition": "non-harmful",
+    "mozPositionDetail": "While this is an important aspect of a robust Spectre-defense, we would like to see a safelist-based approach pursued.",
```
dbaron, line 272: I think it's probably worth explicitly pointing to (or at least naming) the alternative proposal that you're working on.  Since if we're saying we prefer an alternative, we should do in a way that it's clear to readers that we're talking about an alternative that's actually being developed (rather than, say, a proposed change to this spec).

annevk: Would that require some changes to make mozPositionDetail markdown or some such?

dbaron: Currently `mozPositionDetail` accepts arbitrary HTML, so no changes needed!

annevk: Turns out that we should look out for markup more carefully. 😊

PR #337, opened 2020-05-12
Add webserial (harmful)

### LINKED ISSUE #336: Request for position on Serial API

* Specification Title: Serial API
* Specification or proposal URL: https://wicg.github.io/serial/
* Caniuse.com URL (optional): NA
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=webserial
* Mozillians who can provide input (optional): @marcoscaceres 

### Other information
According to https://github.com/WICG/admin/issues/108#issuecomment-626475273 
there is no input from Mozilla. I hope Mozilla does provide input as this spec would be a useful addition to moving the web forward.

martinthomson:
Closes #336.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -676,6 +676,18 @@
   },
   {
     "ciuName": null,
+    "description": "The Serial API provides a way for websites to read and write from a serial device through script. Such an API would bridge the web and the physical world, by allowing documents to communicate with devices such as microcontrollers, 3D printers, and other serial devices. There is also a companion explainer document.",
+    "id": "webserial",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=webserial",
+    "mozPosition": "harmful",
+    "mozPositionDetail": "Devices that offer serial interfaces often expose powerful, low-level functions over the interface with little or no authentication. Exposing that sort of capability to the web without adequate safeguards presents a significant threat to those devices.",
+    "mozPositionIssue": 336,
+    "org": "Proposal",
+    "title": "Serial API",
+    "url": "https://wicg.github.io/serial"
+  },
+  {
+    "ciuName": null,
     "description": "This document specifies the \"SVCB\" and \"HTTPSSVC\" DNS resource record types to facilitate the lookup of information needed to make connections for origin resources, such as for HTTPS URLs. SVCB records allow an origin to be served from multiple network locations, each with associated parameters (such as transport protocol configuration and keying material for encrypting TLS SNI). They also enable aliasing of apex domains, which is not possible with CNAME. The HTTPSSVC DNS RR is a variation of SVCB for HTTPS and HTTP origins. By providing more information to the client before it attempts to establish a connection, these records offer potential benefits to both performance and privacy.",
     "id": "dnsop-svcb-httpssvc",
     "mozBugUrl": null,


annevk:
I somehow overlooked the request to @dbaron. I'll volunteer to make any further changes that might need making here.

marcoscaceres:
Thanks for the speedy turnaround.

mcarbonneaux:
i don't understand why 'harmfull', if you answer the user what serial port the website whant to access et explain the risk at the moment, or add option to activate the fonctionality for who understand what they do!
is very usfull for iot for exemple, for programming esp32 or arduino lilke device is very simple with this api, i hate the fact to be oblige to use chrome to do that....

fabricedesre:
The problem is that the browser can't know what the risk is precisely, because code served over https has no integrity guarantees.

So yes, it's useful and it would be very nice to figure out an appropriate security/trust model for code using these kind of apis.

mcarbonneaux:
with url white list and user formal confirmation and selection of the serial port i think there no spécifique risk (plus you choose what you connect on your serial port).

fabricedesre:
Don't take it badly, but you just repeat the same arguments.

> with url white list and user formal confirmation at each usage and selection of the serial port i think there no specific risk (plus you choose what you connect on your serial port).

That doesn't provide any guarantee (except maybe for *.localhost urls). The amount of successful phishing attacks also argues against "the user can check the url". And let's not go into prompt fatigue that lead people to just click whatever prompt they get to just move ahead.

> no more than native application, that you not know what are compiled (for binary distribution) in... you relie only on antivirus to be sure and the reputation of site where you download the binary.... and it can containt PAT that wake up in the futur...

The goal is not to be "not worse than native apps", it's to keep the better safety guarantees of the web at large.

> with javascript you can read the code and if the code are executed from url (vs native) you can block this site when is know to be bad site in Phishing and Malware Protection of firefox... i think is more safe than native code because is exxecuted in browser sandbox...

How many times did you read the code before it runs in your browser, seriously? And unfortunately Safe Browsing lists don't protect against a server getting attacked to serve malicious content. About the sandbox, the whole point is that these apis move the goal posts about what is sandboxed.

mcarbonneaux:
> That doesn't provide any guarantee (except maybe for *.localhost urls). The amount of successful phishing attacks also argues against "the user can check the url". And let's not go into prompt fatigue that lead people to just click whatever prompt they get to just move ahead.

On the Internet and in computers there are no absolute guarantees.

But this API makes it possible to develop new Web things more securely than on a native application, and i think that's a good thing.

And let's not go into a logic that leads people to migrate to chrome/safari, because it's not perfectly safe. And I think we can find a happy medium, which allows us to have the possibility of having this functionality with all the mechanisms necessary for it to be secure. That's just what I'm saying...

Like for example only activating the API from the "about:config".
is to set up an explicit whitelist to only allow this option on this list.

That this list is only in "about:config", so that only the informed user can use it...

Then if the user goes to one of these urls and is asked on which serial port he should use the API, there is almost no risk of him doing it without knowing what he is doing...

In this context, this allows the informed user to use this functionality in a sufficiently secure manner and does not open a security risk for the majority of others...

> The goal is not to be "not worse than native apps", it's to keep the better safety guarantees of the web at large.

The purpose of adding the serial interface in the browser is precisely to have more secure access to this kind of interface than in the form of a native application. And if done well it will not make the web more insecure.

> How many times did you read the code before it runs in your browser, seriously? And unfortunately Safe Browsing lists don't protect against a server getting attacked to serve malicious content. About the sandbox, the whole point is that these apis move the goal posts about what is sandboxed.

I say that because it's easier to audit readable code than binary. This does not mean that the labda person will do it. But in general it is easier to control for those who know how to do it. So this leads to greater security in general. Because at some point someone will look and warn Phishing and Malware Protection from firefox, and immediately all firefox users will know about it. it is much faster in terms of reaction than an antivirus.

PR #341, opened 2020-05-19
Add position on WebGPU (important)

### LINKED ISSUE #239: WebGPU

* Specification Title: WebGPU
* Specification or proposal URL: https://github.com/gpuweb/gpuweb
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1602129
* Mozillians who can provide input (optional): @kvark, @jdashg

### Other information

WebGPU spec has issues mentioned in https://github.com/gpuweb/gpuweb/issues/534 . Despite that one of the spec repo maintainers behaves like if they are by design. It may make sense to consider this spec `harmful` and not to implement at all. If Google prefers to implement it, it should be a problem of users of their browser (and its forks) only.

kvark:
Closes #239

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -903,6 +903,18 @@
     "url": "https://github.com/WICG/web-codecs"
   },
   {
+    "ciuName": "webgpu",
+    "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
+    "id": "webgpu",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
+    "mozPosition": "important",
+    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. WebGPU can scale better to the capabilities of GPUs with less overhead for users because it's modelled after the intersection of the modern native GPU APIs, so allows developers to express the GPU workloads more explicitly.",
+    "mozPositionIssue": 239,
+    "org": "Proposal",
+    "title": "Web GPU API",
+    "url": "https://gpuweb.github.io/gpuweb/"
+  },
+  {
     "ciuName": "midi",
     "description": "This specification defines an API supporting the MIDI (Musical Instrument Digital Interface) protocol, enabling web applications to enumerate and select MIDI input and output devices on the client system and send and receive MIDI messages. It is intended to enable non-music MIDI applications as well as music ones, by providing low-level access to the MIDI devices available on the users' systems.",
     "id": "webmidi",
@@ -1069,17 +1081,5 @@
     "org": "WHATWG",
     "title": "timezonechange event",
     "url": "https://github.com/whatwg/html/pull/3047"
-  },
-  {
-    "ciuName": "webgpu",
-    "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
-    "id": "webgpu",
-    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
-    "mozPosition": "important",
-    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. WebGPU can scale better to the capabilities of GPUs with less overhead for users because it's modelled after the intersection of the modern native GPU APIs, so it allows developers to express the GPU workloads more explicitly.",
-    "mozPositionIssue": 239,
-    "org": "proposal",
-    "title": "Web GPU API",
-    "url": "https://gpuweb.github.io/gpuweb/"
   }
 ]
--- a/activities.json
+++ b/activities.json
@@ -1073,12 +1073,13 @@
   {
     "ciuName": "webgpu",
     "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
+    "id": "webgpu",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
     "mozPosition": "important",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -1069,5 +1069,16 @@
     "org": "WHATWG",
     "title": "timezonechange event",
     "url": "https://github.com/whatwg/html/pull/3047"
+  },
+  {
+    "ciuName": "webgpu",
+    "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
+    "mozPosition": "important",
+    "mozPositionDetail": "We believe that WebGPU is a key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. It's an API that is future-looking and can scale better to the capabilities of the GPUs, with less overhead for the users. Firefox Nightly has a prototype implementation behind a pref that was able to run the WebGPU examples as of April 2020.",
```
annevk, line 1078:
```suggestion
    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. It's an API that is future-looking and can scale better to the capabilities of GPUs, with less overhead for users. Firefox Nightly has a prototype implementation behind a pref that was able to run the WebGPU examples as of April 2020.",
```

activities.json
```diff
@@ -1069,5 +1069,16 @@
     "org": "WHATWG",
     "title": "timezonechange event",
     "url": "https://github.com/whatwg/html/pull/3047"
+  },
+  {
+    "ciuName": "webgpu",
+    "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
+    "mozPosition": "important",
+    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. It's an API that is future-looking and can scale better to the capabilities of GPUs, with less overhead for users. Firefox Nightly has a prototype implementation behind a pref that was able to run the WebGPU examples as of April 2020.",
```
dbaron, line 1078: A few comments here:

First, we've mostly been trying to keep standards-positions *separate* from Firefox product planning.  So to be consistent with that -- and the way we've phrased other positions, I'd probably drop the last sentence here.

Second, I worry that the remaining sentences sound a little more like marketing for the spec than the voice we usually use in positions.  I'd perhaps drop the "It's an API that is future looking and" part, and maybe expand a little bit (but not too much -- maybe a sentence) on why it better fits the capabilities of GPUs and has less overhead?

activities.json
```diff
@@ -1069,5 +1069,16 @@
     "org": "WHATWG",
     "title": "timezonechange event",
     "url": "https://github.com/whatwg/html/pull/3047"
+  },
+  {
+    "ciuName": "webgpu",
+    "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
```
dbaron, line 1075: There's no `id` in this entry, and it needs one.  (Not sure what you used to make it, but the current `activities.py` shouldn't have done that.)  Probably `webgpu` or `gpuweb`.

activities.json
kvark: That sounds reasonable. I partially rewrote the description, please take another look!

activities.json
kvark: I wrote it by hand. Let me know if I should start over using `activities.py` instead. The "id" is now added.

dbaron: You don't need to start over -- but please check if `./activities.py validate` shows any errors?

activities.json
```diff
@@ -1073,12 +1073,13 @@
   {
     "ciuName": "webgpu",
     "description": "This specification describes support for accessing modern 3D graphics and GPU computation capabilities on the Web.",
+    "id": "webgpu",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1602129",
     "mozPosition": "important",
-    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. It's an API that is future-looking and can scale better to the capabilities of GPUs, with less overhead for users. Firefox Nightly has a prototype implementation behind a pref that was able to run the WebGPU examples as of April 2020.",
+    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. WebGPU can scale better to the capabilities of GPUs with less overhead for users because it's modelled after the intersection of the modern native GPU APIs, allows the users to express the GPU workloads more explicitly.",
```
dbaron, line 1079:
```suggestion
    "mozPositionDetail": "We believe that WebGPU is key to enable more interactive and content-rich applications on the Web than currently possible with WebGL. WebGPU can scale better to the capabilities of GPUs with less overhead for users because it's modelled after the intersection of the modern native GPU APIs, so it allows developers to express the GPU workloads more explicitly.",
```

activities.json
kvark: Will do! Any reason to not run this on CI for every commit?

dbaron: It does run on commits, but I don't think we set it up to run on PRs.  Maybe I should figure out how to do that...

bholley:
Did the document discussed at https://github.com/mozilla/standards-positions/issues/239#issuecomment-571354262 ever get produced?

dbaron:
Sorry, @bholley reminds me that we did [promise](https://github.com/mozilla/standards-positions/issues/239#issuecomment-571721002) further discussion before finalizing this, so I've reverted this for now.

kvark:
That's totally reasonable, thank you!
I'll see if @jgilbert has any draft and start working on it.

kvark:
@bholley actually, I wrote the [security considerations](https://gpuweb.github.io/gpuweb/#security) section of the spec. Does it leave any of the concerns unanswered?

annevk:
To make this a tad less opaque for those following along, https://github.com/gpuweb/gpuweb/pull/771#issuecomment-631496586 has some relevant discussion on the remaining security questions.

PR #345, opened 2020-05-23
Add portals (defer).

### LINKED ISSUE #157: HTML portal element (Portals)

* Specification Title: Portals
* Specification or proposal URL: https://wicg.github.io/portals/
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): @annevk @bzbarsky @hsivonen 

### Other information
This is being promoted recently: https://web.dev/hands-on-portals

I believe bz had [concerns regarding fallback](https://groups.google.com/a/chromium.org/d/msg/blink-dev/SgsbpO08AeI/W6pI3rlNBgAJ) when this was brought up in the [blink-dev intent to implement](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/SgsbpO08AeI).

dbaron:
Closes #157.


rolivo33:
You guys are a bunch of mozilla shit, good thing they're going to go away soon.

rolivo33:
They are an evil to this world, terrible the damage they do. Declaring each technology as harmful just because they can't compete with chrome shows that they are useless.

activities.json
```diff
@@ -602,6 +602,18 @@
...
+    "ciuName": null,
+    "description": "This specification defines a mechanism that allows for rendering of, and seamless navigation to, embedded content.",
+    "id": "portals",
+    "mozBugUrl": null,
+    "mozPosition": "harmful",
+    "mozPositionDetail": "The Portals proposal introduces a significant amount of complexity to the web without either (a) clearly explaining how it differs and doesn't differ from the iframe element or (b) a serious attempt to specify how it interacts with the large number of existing web features (including aspects of the origin model) many of which are critical for security and how it interacts with existing privacy features in web browsers (such as those that depend on partitioning of storage based on origins).  We believe the proposal as it exists today is harmful, but would be willing to reconsider given a more clearly motivated and clearly specified proposal.",
+    "mozPositionIssue": 157,
+    "org": "Proposal",
+    "title": "Portals",
+    "url": "https://wicg.github.io/portals"
```
annevk, line 615:
```suggestion
    "url": "https://wicg.github.io/portals/"
```
I think we decided to add slashes here to avoid redirects?

domenic:
Hey folks!

I wanted to chime in here to express a bit of surprise at this position. Although I appreciate the open-ended sentence at the end, I'm not sure what to make of the rest. The portals proposal has [extensive discussion](https://github.com/WICG/portals#summary-of-differences-between-portals-and-iframes) of the differences between portals and iframes. And, at least the current draft fully specifies the interaction with all web platform features except session history. (Well, it actually fully specifies the interaction with session history, by using the top-level browsing context concept. But this creates too much implementer work for figuring out how to map session history to UI elements such as the back button, so we've had a long-standing acknowledged issue to detail that mapping in a way that makes it easier for implementers to give their users a good user experience.) Certainly, there is no confusion with the origin model in the current draft; it's not modified at all. Finally, although judgments of what are "significant" are subjective, I was under the impression from our meeting with @annevk in Berlin that we'd managed to keep the complexity pretty low, at least in his estimation. The spec is, after all, pretty short.

All that said! The real reason that I'd push for a "defer" instead of "harmful" is that the portals proposal is undergoing significant changes, in responses to the shifting tides over the last year or so in terms of [storage partitioning efforts](https://privacycg.github.io/storage-partitioning/) in browsers. As such, the current draft does not reflect what we intend to implement in Chromium. That's more captured in the explainer. And the explainer itself has a lot of gaps: if you squint at the various "TODO" sections I've added, you can get a sense of where we are heading, but it's not something I'd really encourage anyone to spend time on; we know there's work to do in laying out the proposed modifications to the storage restrictions and partitioning model, permissions, and rendering in a digestible format, first as an explainer, and then as a spec. (These will, unavoidably, increase the complexity of the spec a bit, perhaps even to "significant" levels, but we think this will be worth it for the privacy benefits.)

So, if it's possible under your process, we'd love to withdraw the current portals proposal from consideration. It's too much under-construction at the moment to be fairly judged. We're hopeful that the direction we're heading will make it more interesting to Mozilla. (That direction is, roughly, laser-focused on privacy-preserving prerendering, including accommodations for modern storage partitioning efforts.) It's certainly up to you whether you'd want to mark portals as "harmful" in the meantime, but I think it would be a bit strange, as the position is being taken on an older draft of the spec/explainer, and as mentioned above, seems to have some inaccuracies in how it read that spec/explainer.

PR #348, opened 2020-05-26
Update title, url, and org of Permissions Policy (formerly Feature Policy)

dbaron:
(note that the sorting by title means the entry moves)

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -337,18 +337,6 @@
   },
   {
     "ciuName": null,
-    "description": "This specification defines a mechanism that allows developers to selectively enable and disable use of various browser features and APIs.",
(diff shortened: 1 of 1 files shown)


martinthomson:
I would support a change to "important" also.

The URL of the spec remains the same, which is awkward.

dbaron:
BTW, discussion about changing to `important` should probably go in #24.

clelland:
Re: https://github.com/mozilla/standards-positions/pull/348#issuecomment-634479133

Looking at this late -- the URL has changed (though the old one redirects) and is now canonically https://w3c.github.io/webappsec-permissions-policy/

PR #352, opened 2020-06-01
Stop removing trailing slashes from URLs, to save redirects.

dbaron:
To match the change in #319, we should stop removing the trailing
slashes from URLs; those URLs will generally redirect to re-add the
trailing slash.


martinthomson:
I had no idea we were doing that.  Why would we do that?

PR #353, opened 2020-06-01
Add keyboard lock (harmful).

### LINKED ISSUE #196: Keyboard Lock

* Specification Title: Keyboard Lock
* Specification or proposal URL: https://github.com/WICG/keyboard-lock/
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=700123

### Other information
A popular game streaming service [has indicated](https://old.reddit.com/r/firefox/comments/chrrme/this_is_kinda_getting_really_old_now/euylx5c/) there is value in this feature for their users.

dbaron:
This adds it as a harmful position, although it could also be written as
a defer position based on the feedback in #196.

Closes #196.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -471,6 +471,18 @@
     "url": "https://tools.ietf.org/html/draft-ietf-httpbis-client-hints"
   },
   {
+    "ciuName": "mdn-api_keyboard_lock",
+    "description": "This specification defines an API that allows websites to capture keys that are normally reserved by the underlying host operating system. It is intended to be used by web applications that provide a fullscreen immersive experience (like games or remote access apps).",
+    "id": "keyboard-lock",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=700123",
+    "mozPosition": "harmful",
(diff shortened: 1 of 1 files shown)


tantek (approved):
Agreed. The UX concerns noted in Issue #196 are sufficient to mark the current proposal as harmful. Significant changes to address those concerns would be sufficient to bring this back for re-evaluation.

bholley:
We're doing some thinking about the right way to approach this API and the associated use-cases, will circle back.

martinthomson:
To be more clear about this, we think that this is a reasonable use case, but it might be a poor implementation of that use case.  It might be better to defer for now.

adamroach:
> To be more clear about this, we think that this is a reasonable use case, but it might be a poor implementation of that use case. It might be better to defer for now.

I'm pretty hesitant to take the "good use case, harmful proposal" situation and call it anything other than harmful. Most of our "harmful" designations could be described that way. We've seen other positions in this repository construed completely without nuance in other venues, so I think it's important that we mark things clearly as "harmful" if the currently written-down mechanism is harmful.

tantek (approved):
This: "mark things clearly as "harmful" if the currently written-down mechanism is harmful". 

IMO merge it. It's worth getting on the record the current state, regardless of what the potential state might be in some theoretical optimistic future.

adamroach (approved):
This text looks good. Ship it.

martinthomson (approved):

martinthomson:
The team that is responsible for this has this on their backlog; let's hold this until we collect their input.

zcorpan:
I quoted the rationale in this PR in https://github.com/mozilla/standards-positions/issues/196#issuecomment-4251914260

Let's close this. Let me know if you think the dashboard should include the rationale, then I can make a new PR.

PR #356, opened 2020-06-04
Update expand/collapse UI

### LINKED ISSUE #354: make expand/collapse nature of entries more obvious
One thing I worry about with the [positions table](https://mozilla.github.io/standards-positions/) is that it's not obvious to readers that the entries expand.  Yet the parts that are visible when you expand them are an important part of the message we're trying to convey.

To remedy this, I'd like to make roughly the following changes:
* add expand/collapse triangles at the left edge, replacing (I think) the "Details" column and icon
  * probably this requires moving the link icon to the right edge, but not entirely sure
* add [Expand All] and [Collapse All] buttons at the top

Thoughts?

dbaron:
This does the minimal set of UI updates described in #354.  I didn't touch the link icon, although I could do so later if we think it's worthwhile.

This is probably intended to be landed as three commits rather than squashed.

Fixes #354.

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,6 +62,12 @@
 
       <div class="col-sm-1"></div>
       <div class="col-sm-10">
+
+        <div>
+          <a class="btn btn-primary" href="javascript:open_all()">Expand All</a>
+          <a class="btn btn-primary" href="javascript:close_all()">Collapse All</a>
+        </div>
+
         <table id="mozPositions" class="table table-hover dataTable">
           <thead>
             <tr>
@@ -260,6 +266,12 @@
           }
         }
 
+        window.open_all = function() {
+          ptable.rows().every( function(rowIndex, tableLoop, rowLoop) {
+            open_row(rowIndex);
+          });
+        }
+
         window.open_row = function(row_selector) {
           var row = ptable.row(row_selector);
           var data = row.data();
@@ -272,6 +284,12 @@
             row.child( extra_data ).show();
             tr.classList.add('shown');
           }
+        }
+
+        window.close_all = function() {
+          ptable.rows().every( function(rowIndex, tableLoop, rowLoop) {
+            close_row(rowIndex);
+          });
         }
 
         window.close_row = function(row_selector) {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -321,7 +340,7 @@ <h3>legend</h3>
 
           // Open the row that is now the target.
           var row = ptable.row( tr );
```
martinthomson, line 342: This could call `open_row(tr)`, right?  There is a little more indirection, but you aren't hitting the "private" API.  And if you do that, you no longer need a separate "private" API, you can just roll _open_row into open_row and _close_row into close_row.

index.html
```diff
@@ -62,14 +62,20 @@ <h1><img src="asset/Mozilla.svg" style="height: 1.04em; vertical-align: -0.23em;
 
       <div class="col-sm-1"></div>
       <div class="col-sm-10">
+
+        <div>
+          <a class="btn btn-primary" href="javascript:open_all()">Expand All</a>
+          <a class="btn btn-primary" href="javascript:close_all()">Collapse All</a>
```
martinthomson, line 68: Do we do javascript: links or addEventListener?  I don't mind, but I generally use the latter when I can.

dbaron: I've generally had a preference for the former because it's more open to multiple methods of activation (mouse/keyboard/etc.).  I think assistive technologies have generally learned to work around cases where people only listen to `click`, but maybe not in all cases, and I think it's still bad for keyboard users.

index.html
dbaron: Ah, true.  I was thinking of `open_row` as a function that took the row index as an argument, but really it takes whatever [`ptable.row()`](https://datatables.net/reference/api/row()) takes as an argument, i.e., a [row selector](https://datatables.net/reference/type/row-selector).  So I can pass the row index in all the other cases and the `<tr>` node in this case.

dbaron: Fixed in the updated version (I force-pushed since I updated all three patches in place, so I can still cleanly push the queue of 3 patches).

PR #357, opened 2020-06-05
Make the cross links between positions on Fragmentions and Scroll To Text Fragment be hyperlinks.

dbaron:


martinthomson:
If this were attacker-controlled input, that would make me very nervous.  Even still, it makes me nervous.  I *guess* this is OK, but it means we aren't using `document.createTextNode()` as we should.

annevk:
Yup: 3e43a1999de7503dcb288ea289607ae683d9bd8d.

PR #358, opened 2020-06-09
Add a section to CONTRIBUTING.md on how Mozilla's subject-matter experts should contribute.

dbaron:
When I've solicited feedback from subject-matter experts within Mozilla, I've been asked a number of questions what sort of feedback is useful and how to provide it.  This attempts to write down the advice I've given so that it's in one place and so that we can get wider agreement on what that advice should be.

I'm tagging a few people explicitly for review, but this is one of those cases when I'm most interested in comments from everybody participating here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,6 +87,25 @@ assessment.
 
 ## Discussing Mozilla's Position on a Web Specification
 
+### For the Broader Community
+
+We welcome discussion members of the wider Mozilla community -- including the public -- but ask
+that it be on-topic, and that it follow [Mozilla's Community Participation
+Guidelines](https://www.mozilla.org/about/governance/policies/participation/).
+
+Specifically, the primary purpose of this repository is to determine Mozilla's postions on
+specifications. That is distinct from the larger Web community's position; the best place to
+advocate for a specification is using the appropriate standards body's discussion mechanisms, not
+here.
+
+So, please focus your comments on bringing new information about a specification; if you want to
+express support for adopting a specification, the best way to do that is using [Github
+reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) on the
+issue or a specific comment.
+
+If an issue becomes overwhelmed by excessive advocacy, comments might be deleted, and the
+issue might be locked.
+
 ### For Mozilla's Subject-Matter Experts
 
 You might be asked for your opinion
@@ -130,21 +149,3 @@ happening in the issue
 and limit the discussion in the pull request
 to the details of making the change to `activities.json`.
 
-### For the Broader Community
-
-We welcome discussion members of the wider Mozilla community -- including the public -- but ask
-that it be on-topic, and that it follow [Mozilla's Community Participation
-Guidelines](https://www.mozilla.org/about/governance/policies/participation/).
-
-Specifically, the primary purpose of this repository is to determine Mozilla's postions on
-specifications. That is distinct from the larger Web community's position; the best place to
-advocate for a specification is using the appropriate standards body's discussion mechanisms, not
-here.
-
-So, please focus your comments on bringing new information about a specification; if you want to
-express support for adopting a specification, the best way to do that is using [Github
-reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) on the
-issue or a specific comment.
-
-If an issue becomes overwhelmed by excessive advocacy, comments might be deleted, and the
-issue might be locked.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
 
 ## Discussing Mozilla's Position on a Web Specification
 
+### For Mozilla's subject-matter experts
```
martinthomson, line 90:
```suggestion
### For Mozilla's Subject-Matter Experts
```

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+welcome to judge that we've come to sufficient consensus in the issue,
+and make a pull request to document that consensus by changing `activities.json`.
+When this happens, we'd like to try to keep the technical discussion
+about what the position is
+happening in the issue
+(so that it stays in one place),
+and limit the discussion in the pull request
+to the details of making the change to `activities.json`.
+
+### For the broader community
```
martinthomson, line 133:
```suggestion
### For the Broader Community
```

CONTRIBUTING.md
martinthomson, line 90: I would move this below the "broader community" section.

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
 
 ## Discussing Mozilla's Position on a Web Specification
 
+### For Mozilla's subject-matter experts
+
+You might be asked for your opinion
+on specifications where you are one of the experts within the Mozilla community,
+or feel like you have valuable contributions to make
+to issues on other specifications even where you haven't explicitly been asked to do so.
+This is the purpose of this repository:
```
marcoscaceres, line 96: Maybe move this to be the first sentence and rephrase? Something like:
 
"The purpose of this repository is to help Mozilla's technical leaders come to a position about new features being added to the Web. As such, you might be asked for your opinion..."

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+to help the relevant people in Mozilla's technical leadership come to a conclusion.
+
+The goal of the discussions in this repository is
+to figure out what we think about a specification.
```
marcoscaceres, line 100:
```suggestion
to figure out what we think about a proposal or specification.
```

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+The key factors to consider are:
+
+* how useful we think the specification is likely to be (or how important the use case it's trying to address is), relative to the likely cost of implementing it and maintaining that implementation
```
marcoscaceres, line 103:
```suggestion
* how useful we think the proposed feature is likely to be (or how important the use case it's trying to address is), relative to the likely cost of implementing it and maintaining that implementation
```

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+* whether there are any critical problems we see in the current proposal.
```
marcoscaceres, line 104:
```suggestion
* whether there are any critical problems we see in the current proposal, particularly around privacy, security, and/or implementability across multiple engines.
```

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+
+It's useful to discuss these in the relevant issue in this repository.
+
+At the same time,
+we should avoid doing things that move
+discussions that should happen in the standards body's normal discussion mechanisms
```
marcoscaceres, line 110: "that move discussions that" sounds a bit odd to me... could that be rephrased in some other way?

CONTRIBUTING.md
```diff
@@ -86,6 +87,51 @@ assessment.
...
+from happening instead in the standards-positions issue.
+When it makes sense to do so,
+it's good to raise issues through the normal issue trackers for the specification.
```
marcoscaceres, line 113:
```suggestion
it's good to raise issues through the normal issue trackers for the proposal.
```

annevk:
Thanks for writing this down!

CONTRIBUTING.md
dbaron: Yeah, I initially wanted to do that except that I felt that "broader community" needed to be relative to some antecedent.  I guess I can just do it anyway.

CONTRIBUTING.md
dbaron: I did a variant of what you suggested.

CONTRIBUTING.md
dbaron: I don't want to push quite that hard towards those areas (since I think it's worth considering others too), so I listed them as examples rather than as "particularly".

CONTRIBUTING.md
dbaron: I at least found "to move discussions that", but didn't see anything better.

PR #366, opened 2020-06-12
Fix typo in Bluetooth position

martinthomson:


activities.json
```diff
@@ -968,7 +968,7 @@
...
+    "mozPositionDetail": "This API provides access to the Generic Attribute Profile (GATT) of Bluetooth, which is not the lowest level of access that the specifications allow, but its generic nature make it impossible to clearly. Like <a href=\"#webusb\">WebUSB</a> there is significant uncertainty regarding how well prepared devices are to receive requests from arbitrary sites. The generic nature of the API means that this risk is difficult to manage. The Web Bluetooth CG has opted for weaker protections than those in WebUSB, which require active consent to communicate from the device. This proposal uses a blocklist, which will require constant and active maintenance so that vulnerable devices aren't exploited. This model is unsustainable and presents a significant risk to users and their devices.",
```
annevk, line 971: "but its generic nature make it impossible to clearly" doesn't seem correct.

tantek:
```suggestion
    "mozPositionDetail": "This API provides access to the Generic Attribute Profile (GATT) of Bluetooth, which is not the lowest level of access that the specifications allow, but its generic nature make it impossible to clearly evaluate. Like <a href=\"#webusb\">WebUSB</a> there is significant uncertainty regarding how well prepared devices are to receive requests from arbitrary sites. The generic nature of the API means that this risk is difficult to manage. The Web Bluetooth CG has opted for weaker protections than those in WebUSB, which require active consent to communicate from the device. This proposal uses a blocklist, which will require constant and active maintenance so that vulnerable devices aren't exploited. This model is unsustainable and presents a significant risk to users and their devices.",
```

martinthomson: Clearly my brain is not functioning.

activities.json
```diff
@@ -968,7 +968,7 @@
...
+    "mozPositionDetail": "This API provides access to the Generic Attribute Profile (GATT) of Bluetooth, which is not the lowest level of access that the specifications allow, but its generic nature make it impossible to clearly evaluate. Like <a href=\"#webusb\">WebUSB</a> there is significant uncertainty regarding how well prepared devices are to receive requests from arbitrary sites. The generic nature of the API means that this risk is difficult to manage. The Web Bluetooth CG has opted for weaker protections than those in WebUSB, which require active consent to communicate from the device. This proposal uses a blocklist, which will require constant and active maintenance so that vulnerable devices aren't exploited. This model is unsustainable and presents a significant risk to users and their devices.",
```
annevk, line 971:
```suggestion
    "mozPositionDetail": "This API provides access to the Generic Attribute Profile (GATT) of Bluetooth, which is not the lowest level of access that the specifications allow, but its generic nature makes it impossible to clearly evaluate. Like <a href=\"#webusb\">WebUSB</a> there is significant uncertainty regarding how well prepared devices are to receive requests from arbitrary sites. The generic nature of the API means that this risk is difficult to manage. The Web Bluetooth CG has opted for weaker protections than those in WebUSB, which require active consent to communicate from the device. This proposal uses a blocklist, which will require constant and active maintenance so that vulnerable devices aren't exploited. This model is unsustainable and presents a significant risk to users and their devices.",
```
make it impossible -> makes it impossible

Issue #93, opened 2018-06-21
CSS Houdini


flaki:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: CSS Houdini family features (Paint API, Typed OM, Layout API etc.)
* Specification or proposal URL: [CSS Painting API Level 1](https://drafts.css-houdini.org/css-paint-api/)
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1302328
* Mozillians who can provide input (optional): @tantek, Andrew Overholt

### Other information
* [Intent to Experiment: CSS Houdini Paint API Level 1](https://groups.google.com/forum/#!topic/mozilla.dev.platform/uI5bHZil2KU)
* [CSS/Houdini - Mozilla Wiki](https://wiki.mozilla.org/CSS/Houdini)
* [**META** Implement Houdini CSS Painting API Level 1 spec](https://bugzilla.mozilla.org/show_bug.cgi?id=1302328)
* [**META** Implement Houdini CSS Layout API Level 1 spec](https://bugzilla.mozilla.org/show_bug.cgi?id=1302337)
* [**META** implement Houdini CSS Properties and Values API](https://bugzilla.mozilla.org/show_bug.cgi?id=1273706)
* [**META** Implement CSS Houdini "Worklets Level 1" spec](https://bugzilla.mozilla.org/show_bug.cgi?id=1315239)
* [Houdini "Worklets Level 1" prototype implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1290021)
* [Is Houdini ready yet?](https://ishoudinireadyyet.com/)

Under consideration: _Worklets Level 1, CSS Paint API Level 1, CSS Layout API Level 1_  
Currently being implemented: _CSS Typed OM Level 1, CSS Properties and Values API Level 1_  
— according to [the Mozilla Wiki on DOM/Houdini](https://wiki.mozilla.org/CSS/Houdini)

> _“Worklets is one of the Houdini specs we are "considering" but have no firm implementation plans for. Unlikely for 2018, we can reconsider when doing 2019 CSS priorities.”_
> —  [Tantek Çelik, bugzilla bug #1315239](https://bugzilla.mozilla.org/show_bug.cgi?id=1315239#a47768742_164722)

[...]

PR #368, opened 2020-06-15
WebUSB is worse than I remembered

martinthomson:
The requirement that a device be aware of what is connected to it was
removed from the WebUSB spec.

This removes the unnecessary comparative language.


rektide:
For reference, WebUSB position's PR is #93 .

Apologies for the personal note but, I hope some day we might be willing to consider growing the web. It asks a lot, but I'd love if it these positions included some idea for how or what compromises might shift Mozilla's position. These denials feel total.

PR #369, opened 2020-06-17
Add position for the inert attribute.

emilio:
Mark it as worth-prototyping as per this comment[1] and following.

Closes #174, let's continue the more technical discussion somewhere
else, like in the specification issue.

[1]: https://github.com/mozilla/standards-positions/issues/174#issuecomment-590227139

PR #371, opened 2020-06-18
Add position for the inert attribute.

### LINKED ISSUE #174: HTML: inert attribute

* Specification or proposal URL: https://github.com/whatwg/html/pull/4288 & https://github.com/WICG/inert/blob/master/explainer.md
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 

### Other information

The inert attribute allows for marking a section of a document as not being able to receive UI events. Those will go to the nearest ancestor instead.

emilio:
Mark it as worth-prototyping as per this comment[1] and following.

Closes #174, let's continue the more technical discussion somewhere
else, like in https://github.com/whatwg/html/issues/5650.

Same as #369, but with @dbaron's feedback addressed, because apparently
I can't re-open a PR if I have already force-pushed to the same branch.

[1]: https://github.com/mozilla/standards-positions/issues/174#issuecomment-590227139

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1155,6 +1155,18 @@
     "url": "https://html.spec.whatwg.org/multipage/interaction.html#attr-enterkeyhint"
   },
   {
+    "ciuName": "mdn-api_htmlelement_inert",
+    "description": "Allow arbitrary DOM subtrees to become inert",
+    "id": "inert-attr",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=921504",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "A high-level tool for authors to achieve some of these use cases while getting a11y right seems useful",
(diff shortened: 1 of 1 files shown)


dbaron:
>     1. question for the commit in particular: Why the "mdn-api_" prefix on the ciuName? No other entries have that (except the recent "mdn-api_bluetooth").

I think it seems like [caniuse](https://caniuse.com/) is driving its newer entries off of MDN data rather than maintaining its own data.  (This makes me wonder whether we should still be linking to caniuse rather than MDN.)

marcoscaceres:
Generally speaking, caniuse's data is quite unreliable. It concerns itself with blanket "is this technology supported?" instead of the more nuanced data we get from MDN (per attribute/method/spec-section). I'd be in favor or dropping caniuse from our positions entirely. MDN's data is also not perfect, but it's not bad either.

PR #376, opened 2020-06-25
Add Import Attributes (worth prototyping)

### LINKED ISSUE #373: Import Attributes

* Specification Title: Import attributes
* Specification or proposal URL: https://github.com/tc39/proposal-import-attributes
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 

### Other information

Currently Stage 2.

codehag:
closes #373

I was getting an error when running activities.py -- `ERROR: Can't figure out what organisation github.com belongs to!` but we already have a TC39 proposal here. Would it make sense to patch activities.py or should we just do this manually?

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -508,6 +508,18 @@
   },
   {
     "ciuName": null,
+    "description": "The ES Import Attributes and JSON modules proposal adds an inline syntax for module import statements to pass on more information alongside the module specifier, and an initial application for such attributes in supporting JSON modules in a common way across JavaScript environments.",
+    "id": "import-attributes",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1648614",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "This proposal enables the importing of JSON content into a JS module. This mechanism is a prerequisite for HTML/CSS/JSON modules.",
+    "mozPositionIssue": 373,
+    "org": "Ecma",
+    "title": "Import Attributes",
+    "url": "https://tc39.es/proposal-import-attributes/"
+  },
+  {
+    "ciuName": null,
     "description": "This document proposes a few changes to cookies inspired by the properties of the HTTP State Tokens mechanism.  First, cookies should be treated as \"SameSite=Lax\" by default.  Second, cookies that explicitly assert \"SameSite=None\" in order to enable cross-site delivery should also be marked as \"Secure\".  Third, same-site should take the scheme of the sites into account.  Fourth, cookies should respect schemes.  Fifth, cookies associated with non-secure schemes should be removed at the end of a user's session.  Sixth, the definition of a session should be tightened.",
     "id": "cookie-incrementalism",
     "mozBugUrl": null,


activities.json
```diff
@@ -506,6 +506,18 @@
     "title": "HTTP Client Hints",
     "url": "https://tools.ietf.org/html/draft-ietf-httpbis-client-hints"
   },
+  {
+    "ciuName": null,
+    "description": "The ES Import Attributes and JSON modules proposal adds an inline syntax for module import statements to pass on more information alongside the module specifier, and an initial application for such attributes in supporting JSON modules in a common way across JavaScript environments.",
+    "id": "import-attributes",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1648614",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "This proposal enables importing non-js files into a JS module, specifically in this case JSON. Without this mechanism, it wouldn't be possible to do this in a way that isn't open to midflight changing of the file type, which could then be inteereepreetteed as JS. The proposal is necessary to move forward on HTML/CSS/JSON modules. We consider it important that this is defined inbound rather than out of bound.",
```
dbaron, line 515:
```suggestion
    "mozPositionDetail": "This proposal enables importing non-js files into a JS module, specifically in this case JSON. Without this mechanism, it wouldn't be possible to do this in a way that isn't open to midflight changing of the file type, which could then be interpreted as JS. The proposal is necessary to move forward on HTML/CSS/JSON modules. We consider it important that this is defined inbound rather than out of bound.",
```

dbaron, line 515: It's not clear to me what "inbound" and "out of bound" mean.  Might it be worth making this clearer?

tilgovi: It should probably say "in band" and "out of band". Even with that change, it may still need more clarity.

activities.json
```diff
@@ -506,6 +506,18 @@
...
+    "mozPositionIssue": 373,
+    "org": "Ecma",
+    "title": "Import Attributes",
+    "url": "https://github.com/tc39/proposal-import-attributes"
```
martinthomson, line 519: Don't we normally cite the github.io/spec URL?  In which case it would be 
```suggestion
    "url": "https://tc39.es/proposal-import-attributes/"
```

The reason that you got a weird message from the python script is that it tried to infer some things, but didn't know about this URL.

We probably need to switch tc39.github.io to tc39.es now though (and tweak BigInt).

activities.json
martinthomson: It seems like this might be too much detail: "This proposal enables the importing of JSON content into a JS module. This mechanism is a prerequisite for HTML/CSS/JSON modules."

The point about security is probably OK to skip, and unless there is something fundamental in the disagreement we have regarding in-band vs. out-of-band, then I would skip that alss.  The security advantage seems like that this is an innate property of the proposal, so I'm not sure why it would be called out in this way.  If you want to say something then I would suggest "Explicit indication of file types as proposed has the advantage of avoiding content type substitution attacks."  or something like that.

codehag: Ok, I have updated it to reflect this.

PR #380, opened 2020-07-01
Switch from bower to npm, preserving current versions of dependencies

dbaron:
This switches package management from bower to npm while preserving current versions of dependencies.  (It uses a github version for DataTables since npm doesn't happen to have the exact version we were using, but it has others.)

The changes are, I think:

* creation of the toplevel `package.json` file
* moving most (but not all) things from `components` to `node_modules` (is there a way to avoid this?) (**and corresponding change to `index.html`**)
* various packaging files (e.g., `package.json`, `bower.json`, etc.) inside `components`/`node_modules` added/removed/changed
* a bunch of unneeded files in bootstrap and datatables no longer included
* ~import of the entire datatables plugins package instead of the manual single-file import that I did in a6399166fe9108c88b4b60d497d948d79579baa1, and rename of its directory from `datatables-plugins` to `datatables.net-plugins` (**and corresponding change to `index.html`**)~

... although there are a few other things that appear to be changes but that I concluded are artifacts of git's matching of files.

I think this should make it easier to upgrade stuff.  (I couldn't get bower to work the last time I tried, hence that manual import of the datatables plugin, and it's also [deprecated](https://snyk.io/blog/bower-is-dead/).)

Things seem to work fine when I test locally.


dbaron:
I guess I'm not crazy about having the various HTML files in `datatables.net-plugins` in the repository and thus served on `mozilla.github.io`.  They do load some script from other origins.  Not sure what the best way to avoid that is, though...

dbaron:
Maybe I should just leave the manual import of the one piece of datatables-plugins as it was, and not move that over to the npm configuration?

dbaron:
OK, I did exactly that -- left the partial import of the single datatables plugin as it was, rather than pulling in a rather large project with a bunch of large html files.

nschonni:
What about just using CDN versions?
- https://getbootstrap.com/docs/3.3/getting-started/
- httsp://cdn.datatables.net/1.10.15/css/jquery.dataTables.min.css
- httsp://cdn.datatables.net/1.10.21/js/jquery.dataTables.min.js

mnot:
Using offsite libraries will require a separate connection and incur potential congestion control conflicts. It's better to leave them onsite.

node_modules/datatables/package.json
```diff
@@ -0,0 +1,99 @@
+{
+  "_args": [
+    [
+      "datatables@github:datatables/DataTables#1.10.15",
+      "/home/dbaron/builds/standards-positions"
```
martinthomson, line 5: Is this something that you are comfortable with?

dbaron: I've pasted enough gdb stack traces into bugzilla that I don't think that aspect of my filesystem structure is a secret, if that's what you're asking.

Should have have invoked `npm` in some other way to avoid that?

dbaron:
I had to push 5ac8d1a as a followup to fix the site.

PR #265, opened 2020-02-06
Defer Privacy Pass and Trust Token positions

martinthomson:
Closes #261.
Closes #262.

PR #384, opened 2020-07-02
Add missed issue numbers.

dbaron:
I think these should have been added in #265.


martinthomson:
My bad.  Thanks for fixing the validation.

PR #385, opened 2020-07-02
Make validation require the mozPositionIssue.

dbaron:
This needs to land after #384.


dbaron:
The test is failing just like it should (and it's good to see that), but I promise it'll pass when #384 merges. :-)

PR #392, opened 2020-07-13
rename import attributes to import conditions

xtuc:
The proposal was renamed to import-conditions, I don't know if it's supposed to be edited here.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -521,7 +521,7 @@
   {
     "ciuName": null,
     "description": "The Import Conditions and JSON modules proposal adds an inline syntax for module import statements to pass on more information alongside the module specifier, and an initial application for such attributes in supporting JSON modules in a common way across JavaScript environments.",
-    "id": "import-conditions",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -520,15 +520,15 @@
   },
   {
     "ciuName": null,
-    "description": "The ES Import Attributes and JSON modules proposal adds an inline syntax for module import statements to pass on more information alongside the module specifier, and an initial application for such attributes in supporting JSON modules in a common way across JavaScript environments.",
-    "id": "import-attributes",
+    "description": "The Import Conditions and JSON modules proposal adds an inline syntax for module import statements to pass on more information alongside the module specifier, and an initial application for such attributes in supporting JSON modules in a common way across JavaScript environments.",
+    "id": "import-conditions",
```
dbaron, line 524:
```suggestion
    "id": "import-attributes",
```

I'd prefer to leave the ID as-is since people may have links to it.

Issue #393, opened 2020-07-13
Media Queries Level 5: prefers-reduced-motion, prefers-contrast, prefers-color-scheme


tantek:
## Request for Mozilla Position on an Emerging Web Specification

  * Specification Title: Media Queries Level 5: User Preference Media Features
  * Specification or proposal URL: https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion
  * Specification or proposal URL: https://drafts.csswg.org/mediaqueries-5/#prefers-contrast
  * Specification or proposal URL: https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme
  * Caniuse.com URL (optional):
  * Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1434491
  * Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1506364
  * Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1494034
  * Mozillians who can provide input (optional): @heycam, @dbaron
  * W3C Tag Design Review issue: none

### Other information

This issue is scoped to three of the new features in Media Queries Level 5, which we have implemented, as part of evaluating [Media Queries Level 5 as a whole](https://github.com/mozilla/standards-positions/issues/389): [prefers-reduced-motion](https://bugzilla.mozilla.org/show_bug.cgi?id=1434491), [prefers-color-scheme](https://bugzilla.mozilla.org/show_bug.cgi?id=1494034), and most recently, [prefers-contrast](https://bugzilla.mozilla.org/show_bug.cgi?id=1506364) (as [announced last week](https://wiki.mozilla.org/Firefox/Roadmap/Updates#2020-07-06)).

These features enhance accessibility features of the platform and thus we should consider a status of "important" for these three as a whole.

Label: w3c.

(Originally published at: https://tantek.com/2020/195/b1/)

martinthomson:
@annevk, you commented on the pull request about the (unexplored) privacy risks of these features.  I see that [the issue](https://github.com/w3c/csswg-drafts/issues/3488) remains open.

So for @tantek, I want to ask whether we are comfortable saying that these are important without fully understanding the implications?

hsivonen:
The previous comment above links to a spec issue, and it seems the spec still has as issue box without having resolved the concerns further.

[...]

PR #394, opened 2020-07-13
defer Media Queries 5 until CR

### LINKED ISSUE #389: CSS Media Queries Level 5

  * Specification Title: Media Queries Level 5
  * Specification or proposal URL: https://drafts.csswg.org/mediaqueries-5/
  * Caniuse.com URL (optional):
  * Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1434491
  * Mozillians who can provide input (optional): @heycam, @dbaron
  * W3C Tag Design Review issue: none

### Other information

We have implemented several of the features in Media Queries Level 5: [prefers-reduced-motion](https://bugzilla.mozilla.org/show_bug.cgi?id=1434491), [prefers-color-scheme](https://bugzilla.mozilla.org/show_bug.cgi?id=1494034), and most recently, [prefers-contrast](https://bugzilla.mozilla.org/show_bug.cgi?id=1506364) (as [announced Monday](https://wiki.mozilla.org/Firefox/Roadmap/Updates#2020-07-06)).

In general Media Queries Level 5 has several accessibility enhancing features and thus we should consider an overall status of "important".

Label: w3c.

(Originally published at: https://tantek.com/2020/191/b2/)

tantek:
Closes #389. We will evaluate specific features individually or related sets of features instead until MQ5 reaches CR, and then we can re-open #389 for re-evaluation.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -580,6 +580,18 @@
   },
   {
     "ciuName": null,
+    "description": "Media Queries Level 5 describes the mechanism and syntax of media queries, media types, and media features. It extends and supersedes the features defined in Media Queries Level 4.",
+    "id": "mediaqueries-5",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1434491",
+    "mozPosition": "defer",
+    "mozPositionDetail": "It's possible the scope of mediaqueries-5 may change over time (either due to additional new features being added or to features being moved between levels), and thus we should defer evaluating it overall until it reaches CR.",
+    "mozPositionIssue": 389,
+    "org": "W3C",
+    "title": "Media Queries Level 5",
+    "url": "https://drafts.csswg.org/mediaqueries-5/"
+  },
+  {
+    "ciuName": null,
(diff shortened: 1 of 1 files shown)


annevk:
Given the third bullet point in https://www.w3.org/2019/Process-20190301/#candidate-rec wouldn't that be a tad to late?

dbaron:
What I was suggesting in #389 was that we should not have "Mediaqueries 5" as an entry in the table at all; instead we should have features at the scope that we would make decisions about whether they're worthwhile or not.  CSS modules, particularly for additions to mature features, tend to be a grab bag of features of totally different value and priority, so I think we should probably avoid trying to prioritize them as a whole.  (Probably when we listed [CSS Grid Layout Module Level 2](https://mozilla.github.io/standards-positions/#subgrid) we should really have listed "subgrid", which is actually what we meant when we marked it as `important`.  It's even in the ID!)

martinthomson:
It might then be better to say "css-accessibility-media-queries" or "css-mq5-accessibility" as the target of this and say "worth prototyping".

dbaron:
Closing in favor of resolving #393 instead, as I suggested above.

Issue #375, opened 2020-06-24
consider replacing caniuse links with MDN links


dbaron:
To follow up on https://github.com/mozilla/standards-positions/pull/371#issuecomment-646286685 and following comments -- I think we should consider replacing the links to [Can I use](https://caniuse.com/) with links to [MDN](https://developer.mozilla.org/).  It seems like a number of the newer caniuse tables are (based on the IDs) powered by MDN data underneath anyway, and it seems like MDN provides more depth of information.

Thoughts?

adamroach:
Given how much more detail MDN provides, this sounds like a net improvement to me. On a quick glance through those few entries with non-null `ciuName` fields, I see straightforward replacements for most of them:
- https://developer.mozilla.org/en-US/docs/Web/API/CSS_Painting_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Browser_compatibility (for "lazy loading attribute")
- https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/Streams_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/SyncManager#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Bluetooth_API#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/USB#Browser_compatibility


WebMIDI is an exception, as there is no parent article, so we'll probably want to support linking to multiple pages:
- https://developer.mozilla.org/en-US/docs/Web/API/MIDIAccess#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/MIDIInput#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/MIDIConnectionEvent#Browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/MIDIInputMap#Browser_compatibility

[...]

PR #396, opened 2020-07-15
Add a way to show links to MDN, and show them for a number of specs.

dbaron:
Add a way to show links to MDN, and show them for a number of specs.
    
This fixes part of #375.  It does the following:
    
 * adds an `mdnUrl` field to `activities.json`, using @adamroach's list in
   https://github.com/mozilla/standards-positions/issues/375#issuecomment-649040392
   which can either be a string or an array of strings (like `mozBugUrl`)
    
 * adds a way to show the MDN icon in the "More Info" column (which I
   commit as an asset rather than picking a glyphicon, and use via CSS
   mask).
    
 * adds code to validate the new field in activities.py (matching
   `mozBugUrl`)
    
 * adds code to add the new field for the `add` verb in `activities.py`

There's also an initial commit to fix something I broke in #385, so this should be rebased rather than squashed.


dbaron:
Oops, I realize I should also make the `activities.py add` code add this field.

activities.json
```diff
@@ -1050,6 +1059,12 @@
     "ciuName": "midi",
     "description": "This specification defines an API supporting the MIDI (Musical Instrument Digital Interface) protocol, enabling web applications to enumerate and select MIDI input and output devices on the client system and send and receive MIDI messages. It is intended to enable non-music MIDI applications as well as music ones, by providing low-level access to the MIDI devices available on the users' systems.",
     "id": "webmidi",
+    "mdnUrl": [
+      "https://developer.mozilla.org/en-US/docs/Web/API/MIDIAccess",
+      "https://developer.mozilla.org/en-US/docs/Web/API/MIDIInput",
+      "https://developer.mozilla.org/en-US/docs/Web/API/MIDIConnectionEvent",
+      "https://developer.mozilla.org/en-US/docs/Web/API/MIDIInputMap"
+    ],
```
martinthomson, line 1067:
```suggestion
    "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/API/MIDIAccess",
```

I think that this is the top-level API, and if it isn't, that's OK.  https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API doesn't exist yet, which would have been better.

PR #398, opened 2020-07-21
Add Generic Sensor API (harmful)

### LINKED ISSUE #35: RFP: Generic Sensors

* Specification Title: Generic Sensor API - and friends...
* Specification or proposal URL: https://www.w3.org/TR/generic-sensor/

### Other information
This has also spawned other sensors that inherit the generic interface - which refactor sensors we already have exposed (and in many cases had to remove) in the browser.

annevk:
Closes #35.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -437,6 +437,19 @@
   },
   {
     "ciuName": null,
+    "description": "This specification defines a framework for exposing sensor data to the Open Web Platform in a consistent way. It does so by defining a blueprint for writing specifications of concrete sensors along with an abstract Sensor interface that can be extended to accommodate different sensor types.",
+    "id": "generic-sensor",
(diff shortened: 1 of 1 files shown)


martinthomson:
The rationale here seems a little lean.  Concentrating on the fact that these are very hard to understand is probably the high-order bit, but it might be good to expand a tiny bit.

We might say that we prefer to find ways to address the underlying use cases that don't rely on providing raw sensor feeds.  Raw sensor feeds have been a reliable source of security and privacy issues.

annevk:
Thanks, I was struggling a bit with how to word this. Let me know what you think.

annevk:
Thanks all. I filed https://github.com/w3c/dap-charter/issues/106.

PR #404, opened 2020-07-23
ids shouldn't have spaces

tantek:
AFAIK, ids shouldn't have spaces. Editorial.


dbaron:
The spec rule appears to be "must not contain any ASCII whitespace".  We should probably make `activities.py validate` check that.

dbaron:
I created #405 for that.

PR #405, opened 2020-07-23
Validate that IDs don't contain whitespace.

dbaron:
This is a followup to #404.

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -18,6 +18,7 @@ import json
 import os
 import re
 import sys
+import string
 from urllib.parse import urlsplit, urlunsplit
 
 try:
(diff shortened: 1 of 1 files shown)


dbaron:
(I was hoping python would have a convenient method to do this sort of thing, but I didn't see one, despite vaguely remembering that there should be one.)

activities.py
```diff
@@ -39,6 +40,10 @@
 JSON_ENCODER = json.JSONEncoder(sort_keys=True, indent=2, separators=(",", ": "))
 
 
+class IdType(object):
+    "indicates a ID attribute."
```
dbaron, line 44:
```suggestion
    "indicates an ID attribute."
```

Issue #101, opened 2018-08-08
MediaStreamTrack Content Hints


alvestrand:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: MediaStreamTrack Content Hints
* Specification or proposal URL: https://w3c.github.io/mst-content-hint/
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): jan-ivar

annevk:
cc @jan-ivar

jan-ivar:
Sorry @nils-ohlmeier for taking a while to collect my thoughts. I have several concerns with this API.

### Weak API

It's an odd control surface. It doesn't feel like a control surface.

This is a JS app telling the browser something, and expecting functionality in return, yet it doesn't appear to provide any normative language that a browser MUST do anything in response, which seems weak. It's only a "hint" and not "type", right? (If I'm mistaken, then the API seems to a bit of a misnomer).

# Action at a distance.

It feels like a control surface masquerading as a property, or a message channel: "To whomever finds this note..."—It's a lateral API attempting to influence (control?) multiple sinks, potentially at the same time. Maybe. Playback? PeerConnection? MediaRecorder? element.captureStream? AudioContext.createMediaStreamSource()?

In WebRTC, we rejected `pc.addStream` over this; it was deemed undesirable to control a PeerConnection implicitly through a MediaStream, and how many tracks were in it.

### Permanence issues

These hints don't make it through a peer connection, web audio, `element.captureStream()` or any other modifications, right?—JS will have to stick these hints back on the track like post-it notes that keep falling off.

### Under-specified

* How should browsers behave if the JS twiddles the bit live over time?
* How does it work with track.clone()? Can each clone have its own value? 

The spec doesn't say.

### Redundancy and confusion over conflicts

There seems to be significant overlap with existing control surface APIs, like `pc.setParameters()` and
`applyConstraints({echoCancellation, noiseSuppression, autoGainControl})`.

This can lead to conflict or at least confusion over net behavior.

This all combines into a bit of a "throw salt over your shoulder" API for me. Line up all the rocks and wave the sticks right, and maybe the planes will land, or at least some will, maybe not all.

I think we're better off closing any functionality gap in existing APIs, instead of introducing a competing API.

nils-ohlmeier:
After looking at @jan-ivar review and some more internal discussion I think we should be saying "defer" to this as we don't consider JS giving hints is the right solution to the problem.

[...]

PR #408, opened 2020-07-24
Add Content Hints (worth prototyping)

jan-ivar:
Closes https://github.com/mozilla/standards-positions/issues/101.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -255,7 +255,7 @@
   {
     "ciuName": null,
     "description": "This specification extends MediaStreamTrack to let web applications provide an optional media-content hint attribute. This helps sinks like RTCPeerConnection or MediaRecorder select encoder parameters and processing algorithms appropriately based on a hint about the nature of the content being consumed without having to examine the actual content.",
-    "id": "content-hints",
+    "id": "mst-content-hint",
     "mdnUrl": null,
     "mozBugUrl": null,
     "mozPosition": "worth prototyping",
--- a/activities.json
+++ b/activities.json
@@ -254,19 +254,6 @@
   },
   {
     "ciuName": null,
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -252,6 +252,19 @@
     "title": "Contact Picker API",
     "url": "https://wicg.github.io/contact-api/spec/"
   },
+  {
+    "ciuName": null,
+    "description": "This specification extends MediaStreamTrack to let web applications provide an optional media-content hint attribute. This helps sinks like RTCPeerConnection or MediaRecorder select encoder parameters and processing algorithms appropriately based on a hint about the nature of the content being consumed without having to examine the actual content.",
+    "id": "content-hints",
```
dbaron, line 258:
```suggestion
    "id": "mst-content-hint",
```

For W3C drafts, I tend to prefer using the TR shortname as the `id` unless there's a good reason not to.  ("content-hints" also sounds a bit more general.)

activities.json
```diff
@@ -252,6 +252,19 @@
...
+    "mdnUrl": null,
+    "mozBugUrl": null,
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "Content Hints is a welcome higher-level abstraction that does not require broad knowledge and tuning of video encoder, audio-processing, and congestion controls directly. Early concerns over lack of specificity around how hints interact with the lower-level controls they complement have been addressed.",
+    "mozPositionIssue": 101,
+    "org": "W3C",
+    "title": "Content Hints",
```
dbaron, line 265:
```suggestion
    "title": "MediaStreamTrack Content Hints",
```

I think it's probably better to use the title that's on the spec here -- it's also much less ambiguous, whereas "Content Hints" could be a lot of different things to somebody familiar with Web technology but not this particular spec.

PR #416, opened 2020-08-05
prompt MDN URL in new issues

tantek:


activities.py
```diff
@@ -306,6 +306,7 @@ def create_issue(self):
 * Specification Title: {title}
 * Specification URL: {url}
 * Caniuse.com URL (optional): {ciuName}
```
martinthomson, line 308: We should drop caniuse, since that is defunct now.  Note that this formats the string using the `{ciuName}` field which was never set.  Similarly, there is no way that we are populating `{mdnUrl}`, so that ends up empty.

tantek: This sounds reasonable for new specs/features moving forward, it's more likely they'll be documented in MDN than caniuse.

Issue #23, opened 2017-09-27
Payment Handler API


marcoscaceres:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Payment Handler API
* Specification or proposal URL: https://w3c.github.io/payment-handler/

### Other information
The Payment Handler API allows a web application to register itself as capable of handling a "payment request". That is, it claims to either be able to settle a monetary transaction or is potentially capable of supplying a requesting website with payment instrument (e.g., a credit card) on behalf of the end-user. 

For example, `wallet.com` could register itself as being able to handle requests for credit card information on behalf of the user. Then merchant.com can say: "I support 'wallet.com', if you'd like to pay with that". Then requests for payment are routed to `wallet.com` (via a service worker), and `wallet.com` responds with either a proof-of-settlement token, or possibly a payment instrument like a  credit card number. 

This allows Payment Handlers to work along side the browser's built in wallet: giving the user a choice of either picking a credit card form the browser's wallet, or letting `wallet.com` provide a token or credit card number (and possibly other things... that's still up for discussion).

This specification already has some Mozilla support (e.g., @adamroach and I have contributed to it) - but wanted to bring it to people's attention while it's still early.

marcoscaceres:
@dbaron, as @stpeter and I are both going to be editing this spec, could you please mark this as `participating`?

dbaron:
That seems reasonable, but I think it's also useful to document here why the spec is important.  In other words: how does it fit into the plan for payments, what benefits does it have, etc.?

(For example, one of the things I'd like to see from the Web Payments work is the ability for trusted browser UI to tell the user that they're paying only amount X, which requires that the user or browser not give the site a credential (e.g., a credit card number) that could be used to get more money than that).  It's not clear to me whether this API is part of the path to that.)

marcoscaceres:
> That seems reasonable, but I think it's also useful to document here why the spec is important. In other words: how does it fit into the plan for payments, what benefits does it have, etc.?

Right now, the Payment Request API only allows either the browser or native applications to handle request for payment. For example, the browser may present a UI to provide a website with credit card information. Unfortunately, the Payment Request API is only empowering a tiny number of proprietary  payment methods: "Apple/Microsoft/Google/Samsung Pay". This greatly centralizes payment solutions to a couple of players, locking out everyone else.  

[...]

PR #420, opened 2020-08-06
Payment Handler API (defer)

marcoscaceres:


annevk:
I thought that we only used `defer` if we decided not to look at something. From the description it seems like we looked at this quite a bit and therefore `non-harmful` seems more appropriate.

tantek (changes requested):
#23 is closed, should this pull request instead be referring to #390? If so, that issue still needs a clear summary rationale in the issue itself which is then reflected here. If not, how do we close #390?

tantek (changes requested):
We need to:
* ✅ re-open https://github.com/mozilla/standards-positions/issues/23
* record the reasoning for a new position of "non-harmful" in that issue, and then 
* update this pull request accordingly, referencing 23 to re-close it.

tantek (approved):
Accurate update.

zcorpan:
I've rebased this to match current conventions. But #23 has recent comments. I'll wait a bit with merging to avoid flip-flopping.

zcorpan:
Let's land this as `defer` for now.

PR #421, opened 2020-08-06
Meta: make add work for non-spec URLs

annevk:
This seems like more useful behavior for the various proposals we have to evaluate.

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -239,11 +239,11 @@ class SpecEntry(object):
         self.orig_url = spec_url
         self.data = {
             "id": u"",
-            "title": None,
+            "title": "",
             "description": None,
             "ciuName": None,
             "org": None,
-            "url": None,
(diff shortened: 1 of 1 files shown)


activities.py
```diff
@@ -239,11 +239,11 @@ def __init__(self, spec_url):
         self.orig_url = spec_url
         self.data = {
             "id": u"",
-            "title": None,
+            "title": "",
             "description": None,
             "ciuName": None,
-            "org": None,
-            "url": None,
+            "org": "Proposal",
```
smaug----, line 245: How is org "Proposal"? Shouldn't it be the place where proposal has been made? Or perhaps this default value could be
"WhatWG|W3C|IETF|etc" to hint what the value should be?

martinthomson:
Thanks.  This will make it easier to add miscellaneous stuff.

PR #432, opened 2020-08-21
use-cases should be elsewhere e.g. Explainers

tantek:


CONTRIBUTING.md
```diff
@@ -103,6 +103,9 @@ specifications. That is distinct from the larger Web community's position; the b
 advocate for a specification is using the appropriate standards body's discussion mechanisms, not
 here.
 
+In particular, use-cases should be documented elsewhere (e.g. blog posts), 
+and cited in the respective Explainers (if any) of applicable proposals, rather than this repo.
```
martinthomson, line 107: I think that I would say something different here.

> We expect that cited resources (such as an explainer or proposed specification) include all the information necessary to understand proposal: the problem that is being addressed or use cases of interest; the general shape of the proposed solution; an analysis of how the proposal addresses the problem; and how the proposal might affect the system as a whole.  Issues should only provide a high-level summary of this information in a couple of sentences at most.

martinthomson: @tantek Does this seem better?

tantek: Yes that helps discourage the additional problem behaviors of links off to random GDocs etc. that we've seen in a few instances. I'll update the patch accordingly

PR #440, opened 2020-09-20
Add some missing external links

atjn:
I noticed that a link to the bug for Cookie Store API was missing, and then while i was at it, decided to quickly go through and add Caniuse links where they were missing :)


annevk (approved):

annevk:
Thanks!

PR #443, opened 2020-09-23
Meta: use GitHub Actions

annevk:
Based on https://github.com/whatwg/whatwg.org/blob/master/.github/workflows/build.yml.

I mostly created this because I couldn't easily click through to the circleci results (you have to login?) and it seems that wouldn't be a problem if we used GitHub directly.

(I don't have enough privileges to fully test this however.)

(In case anyone is wondering about `requirements.txt`, you're not alone: https://github.com/whatwg/whatwg.org/pull/333.)


froydnj:
Github Actions don't work on Mozilla's account, fwiw.

Toppole69 (approved):
- [ ]

.github/workflows/validate.yml
```diff
@@ -0,0 +1,19 @@
...
+    - master
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
```
martinthomson, line 17:
```suggestion
```

The current default appears to be 3.9.1, which works fine.  It's probably OK not to bind strongly to a specific version, our use of python is fairly constrained.

martinthomson (commented):
Apparently we are now able to use actions, so let's give this a go.

martinthomson (approved):

.github/workflows/validate.yml
annevk: I can make this '3.x' per https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#specifying-a-python-version. Is that what you meant?

martinthomson: I was thinking that we could just strike the "with" section entirely.

annevk:
It seems to work, cool. I guess you need to merge it mt as the CircleCI check fails. We should also change the GitHub configuration in the admin panel to require the Validate check and no longer require the CircleCI check (and clean up any other integration remnants).

PR #445, opened 2020-09-30
Top-Level Await (worth prototyping)

### LINKED ISSUE #444: Top-level await in JS

* Specification Title: Top-level await
* Specification or proposal URL: https://github.com/tc39/proposal-top-level-await
* HTML integration PR: https://github.com/whatwg/html/pull/4352
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1519100
* Mozillians who can provide input (optional): @codehag @jorendorff

### Other information

The TC39 proposal advanced to Stage 3 at the meeting in June, 2019. Chrome is planning to ship it soon, likely in 88.

Interestingly, Node has flipped on the feature flag in V8 for top-level await by default, so it is already shipping for the Node ecosystem: https://github.com/nodejs/node/pull/34558

codehag:
Resolves #444

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1060,6 +1060,18 @@
   },
   {
     "ciuName": null,
+    "description": "Top-level await enables modules to act as big async functions: With top-level await, ECMAScript Modules (ESM) can await resources, causing other modules who import them to wait before they start evaluating their body.",
+    "id": "top-level-await",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1519100",
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "An ergonomic way of handling modules that contain top level asynchronous work.",
+    "mozPositionIssue": 444,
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
I always learned that constructors should ensure that object invariants were true from the moment that the object was constructed.  This seems like a good way to guarantee that at a module level.  Thanks for the writeup.

tantek (approved):
Agreed. Consistent with discussion in #444. From that and @martinthomson's review it sounds like this could lean towards "important" in the future but I'll leave that as a follow-up for 444.

codehag:
Yeah I was between "important" and "worth prototyping". So far module code hasn't had as much adoption due to the need to load multiple dependencies and use a url schema. This does not affect scripts or bundled apps, so it's affect may not be so great on the web. The more significant use case is node and embedded. I am happy to change it.

PR #446, opened 2020-10-01
Correct typos

jugglinmike:


activities.json
```diff
@@ -1380,7 +1380,7 @@
   },
   {
     "ciuName": null,
-    "description": "A descriptor for CSS @page rules that changes the orientation of the page in generated PDF output (or similar) without otherwise affecing layout.",
+    "description": "A descriptor for CSS @page rules that changes the orientation of the page in generated PDF output (or similar) without otherwise effecting layout.",
```
martinthomson, line 1383:
```suggestion
    "description": "A descriptor for CSS @page rules that changes the orientation of the page in generated PDF output (or similar) without otherwise affecting layout.",
```
"effecting" would mean "causing" in this context, which was not the intent; what is meant is "altering".

martinthomson (commented):
Thanks, I think that you changed the intent in the last one though.

activities.json
jugglinmike: Thanks! (Leave it to me to make a spelling mistake in a patch intended to correct spelling mistakes.) Should be good now.

PR #449, opened 2020-10-05
update url s/feature/permissions

tantek:
Per https://github.com/mozilla/standards-positions/pull/348#issuecomment-703827454 link directly to webappsec-permissions-policy


martinthomson:
I guess that we're no longer [cool](https://www.w3.org/Provider/Style/URI)

PR #450, opened 2020-10-09
Update Permissions Policy position per recent developments

### LINKED ISSUE #24: RFP: Feature Policy

annevk:
Closes #24.

cc @johannhof

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -799,7 +799,7 @@
     "id": "feature-policy",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1390801",
     "mozPosition": "worth prototyping",
-    "mozPositionDetail": "Mozilla's primary interest in this specification is in the delegation of permissions to third-party contexts, and in particular permissions that might require the user to make a choice. To that end we are willing to prototype and ship the allow attribute, but given this reduction in scope the remaining functionality needs to be reevaluated in terms of naming and applicability.",
+    "mozPositionDetail": "Mozilla's primary interest in this specification is in the delegation of permissions to third-party contexts that are not granted there by default. To that end we have shipped the allow attribute. The recently revised Permissions-Policy header seems worth prototyping and would allow for sites to impose restrictions on themselves. We hope that the API can be folded into the Permissions API and have not evaluated the reporting functionality as of yet.",
     "mozPositionIssue": 24,
     "org": "W3C",
     "title": "Permissions Policy",
--- a/activities.json
+++ b/activities.json
@@ -799,7 +799,7 @@
     "id": "feature-policy",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1390801",
     "mozPosition": "worth prototyping",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -799,7 +799,7 @@
     "id": "feature-policy",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1390801",
     "mozPosition": "worth prototyping",
-    "mozPositionDetail": "Mozilla's primary interest in this specification is in the delegation of permissions to third-party contexts, and in particular permissions that might require the user to make a choice. To that end we are willing to prototype and ship the allow attribute, but given this reduction in scope the remaining functionality needs to be reevaluated in terms of naming and applicability.",
+    "mozPositionDetail": "Mozilla's primary interest in this specification is in the delegation of permissions to third-party contexts that are not granted there by default. To that end we have shipped the allow attribute. The recently revised Permissions-Policy header seems worth prototyping and would allow for sites to impose restrictions on themselves. We hope that the API can be folded into the Permissions API and have not evaluated the reporting functionality as of yet.",
```
johannhof, line 802: Maybe

> We hope that the *JavaScript* API can be folded into the Permissions API

or DOM API or something?

johannhof (approved):
Seems good but I don't think I can merge PRs in this repo :)

activities.json
annevk: Would you add that before both occurrences of API? I guess I could do that, but I'm not sure if we consistently do that.

johannhof: No, I would think that it's clear from context after the first use.

If we consistently associate "API" with DOM/JS API on the page then that's fine. It just took me a little more time to parse the meaning of that sentence because I associated "the API" to be synonymous with "Permissions Policy as a whole", but it's really not that important :)

tantek (changes requested):
Agreed with clarifying "the API" to disambiguate. That's worth capturing.

activities.json
annevk, line 802:
```suggestion
    "mozPositionDetail": "Mozilla's primary interest in this specification is in the delegation of permissions to third-party contexts that are not granted there by default. To that end we have shipped the allow attribute. The recently revised Permissions-Policy header seems worth prototyping and would allow for sites to impose restrictions on themselves. We hope that the JavaScript API can be folded into the Permissions API and have not evaluated the reporting functionality as of yet.",
```

tantek (approved):
This is sufficient detail. For any more detail folks can view #24. Thanks

PR #452, opened 2020-10-22
Add pre-commit hook

martinthomson:
This can be used to avoid embarrassing and easy to fix mistakes by
running validation before allowing submissions.

Run `./pre-commit install` to install it.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# This is a pre-commit hook that validates code formatting.
+#
+# Install this by running the script with an argument of "install",
+# which adds a symlink at .git/hooks/precommit; or run the equivalent:
+#
+# $ ln -s ../../hooks/pre-commit .git/pre-commit
+
+root="$(git rev-parse --show-toplevel 2>/dev/null)"
+
+# Some sanity checking.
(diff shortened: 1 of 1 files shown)


annevk:
Why don't we forbid pushes to master instead and require status checks to pass?

martinthomson:
If we don't enforce these checks, in practice we never push to master (soon: main).  This is more for those who are building the json file locally for pull requests.  It helps to have faster feedback than CI.  That's all.

I forgot to run the check locally myself and this tripped me up, so this is my reaction.

martinthomson:
Oh, I just realized: this isn't a pre-commit hook that will run on this repository on GitHub.  I don't even know if that is possible.  This is for your personal clone.

annevk (approved):
Fair enough. I guess I'd still recommend requiring PRs and requiring them to pass status checks though.

tantek (approved):
Seems fine. I also agree with @annevk on "requiring PRs and requiring them to pass status checks".

martinthomson:
OK, I have tweaked the rules for submissions.  No pushes to master and PRs are required with one approval.  I've kept them loose for now, but we can tighten as we learn what works.

PR #454, opened 2020-10-30
Import Maps (worth prototyping)

### LINKED ISSUE #146: Import maps

* Specification Title: Import maps
* Specification or proposal URL: https://github.com/WICG/import-maps
* Mozillians who can provide input (optional): @lukewagner @tschneidereit @linclark @callahad

This rather fundamentally impacts all fetching infrastructure and it's the way you'd polyfill lack of support for a JavaScript built-in module.

(It doesn't really feel like a primitive however as it's not necessarily available before the first fetch (e.g., one from a `Link` header), though perhaps if we'd delay those that could be made to work.)

codehag:
cc @moztcampbell and @annevk.

Closes #146

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -609,6 +609,18 @@
   },
   {
     "ciuName": null,
+    "description": "Import maps allow web pages to control the behavior of JavaScript imports.",
+    "id": "import-maps",
+    "mozBugUrl": null,
+    "mozPosition": "worth prototyping",
+    "mozPositionDetail": "We support the overal intent of the proposal and consider it worth prototyping. There are a few technical details that may require some care, in particular the relationship between speculative parsing and dynamic import maps.",
+    "mozPositionIssue": 146,
+    "org": "Proposal",
+    "title": "Import Maps",
+    "url": "https://wicg.github.io/import-maps/"
+  },
+  {
+    "ciuName": null,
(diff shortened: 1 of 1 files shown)


tantek (changes requested):
I don't disagree with this summary, however we (perhaps you Yulia?) really should first comment with a more thorough conclusion of why "We support the overall intent", in issue #146 so that the reasoning is captured there, and in particular as a clear reply to Dominic's request for an update from Mozilla. Then we can give it a few days (week?) for any dissent/confirmation to be expressed from Mozilla folks (@annevk ?) and then if that goes well we can merge this.

tantek (approved):
Thanks for the comment on #146! LGTM. Let's get a +1 from @annevk then merge.

annevk (approved):

benwiley4000:
Sorry if this is written somewhere obvious, but is there a way to track the actual progress of this prototype? I have some integration tests for my app can only run with Chrome until the import map feature is available in Firefox and I'd be excited to try it out when an early version is ready.

myakura:
@benwiley4000 

> Sorry if this is written somewhere obvious, but is there a way to track the actual progress of this prototype? I have some integration tests for my app can only run with Chrome until the import map feature is available in Firefox and I'd be excited to try it out when an early version is ready.

you might want to check out https://bugzilla.mozilla.org/show_bug.cgi?id=1688879 , which is linked from the position entry https://mozilla.github.io/standards-positions/#import-maps

PR #476, opened 2021-01-15
Rename default branch

annevk:
@tantek @martinthomson in https://github.com/mozilla/standards-positions/settings/branches you can rename the default branch as per https://github.com/github/renaming#creating-a-seamless-move-for-existing-repositories-.

Once that's done this pull request can be marked as ready for review at which point the bot will hopefully trigger and indicate success. If it does not, a force push or some such might be needed.

I used `grep -r master * .* --exclude-dir={..,.,.git,node_modules}` to find impacted files.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -2,10 +2,10 @@ name: Validate
 on:
   pull_request:
     branches:
-    - master
+    - main
   push:
     branches:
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
I was waiting until the tools were available for this, but I guess they are already.  Thanks.

PR #503, opened 2021-03-17
Added and corrected links to caniuse.com

SebastianZ:
I went through all entries and checked whether there's a reference at caniuse.com. Then I've added the missing ones and corrected the broken links.

Sebastian


annevk (approved):
Thank you, this is great!

PR #545, opened 2021-06-17
Update File System Access (harmful)

### LINKED ISSUE #154: File System Access API

* Specification Title: Native File System API
* Specification or proposal URL: https://wicg.github.io/file-system-access/ (https://github.com/WICG/file-system-access)
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 

### Other information
Based on activity in this [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=853326), Google seems to be actively working towards implementing this [draft spec](https://wicg.github.io/native-file-system/).

This spec is a follow-on to the Writable Files proposal, which Mozilla initially classified as *defer*, but the last comment in the position [issue](https://github.com/mozilla/standards-positions/issues/110) suggested a new issue be opened if there is "new information such that reassessment is warranted".

annevk:
Closes #154 (again).


tantek (dismissed):
I'd drop "large" (s/large subset/subset). Given tendency of our positions to be overly optimistically interpreted by advocates / other implementers, we need to err on the side of being more critical of the harms (which this PR helps a lot with), and reserved about the things we're ok with or like.

martinthomson (approved):

rolivo3:
Sad

activities.json
```diff
@@ -452,8 +452,8 @@
...
     "id": "native-file-system",
     "mozBugUrl": null,
-    "mozPosition": "defer",
-    "mozPositionDetail": "The ability to read and write from the filesystem is potentially very dangerous. We will need to carefully consider any solution in light of the security and privacy implications. We recognize that the spec authors take this issue seriously, but we are concerned that any solution will increase the risk of security incidents more than we are willing to tolerate. Right now, there isn't enough detail in the specification to make an assessment of these risks, so we will defer our decision until we have more information.",
+    "mozPosition": "harmful",
+    "mozPositionDetail": "There's a large subset of this API we're quite enthusiastic about (in particular providing a read/write API for files and directories as alternative storage endpoint), but it is wrapped together with aspects for which we do not think meaningful end user consent is possible to obtain (in particular cross-site access to the end user's local file system). Overall we consider this harmful therefore, but Mozilla could be supportive of parts, provided this were segmented better.",
```
annevk, line 456:
```suggestion
    "mozPositionDetail": "There's a subset of this API we're quite enthusiastic about (in particular providing a read/write API for files and directories as alternative storage endpoint), but it is wrapped together with aspects for which we do not think meaningful end user consent is possible to obtain (in particular cross-site access to the end user's local file system). Overall we consider this harmful therefore, but Mozilla could be supportive of parts, provided this were segmented better.",
```

jimmywarting:
Could I ask if #154 maybe be unlocked again, please?
It can remain closed. But i hope it have cooled down enough over a couple of years that it isn't longer overwhelmed and some have moved their focus elsewhere. 

it would be sad if FS access never gets to see the light of day. 
So i hope ppl can remain to the topic and discuss ways to:
- Making it safe(er). and what Mozilla/Firefox could do to implement it (maybe just parts of it)
- Discuss what "FS access API" can do to make Mozilla reconsider.

Maybe edit the original message with a footer describing what is okey to post and what's not.
Maybe put in what it's about the "FS access" that is considered harmful.

PR #579, opened 2021-09-17
UA Client Hints is less than harmful

martinthomson:
After talking to Google, we learned that the additional information they
are including here is not new (as it first appeared).  On balance, that
makes this less than outright harmful.  The goal is only to replicate
information that User-Agent provides in a more structured form, with the
"high entropy" elements gated behind an Accept-CH gate.

That doesn't mean that we are fully supportive of the proposal as it
stands.

We like the idea that User-Agent might be frozen.  This is
something we have already undertaken in Firefox with some success.  This
is something we would like to see, in particular for some of the "high
entropy" elements that are added in the proposal.

We are also positive regarding a JS API that makes the information
available in a less opaque and brittle way.  In particular, giving
browsers a path to rejecting access to information where the use/privacy
trade-off is not favourable (like the WOW bit on Windows systems).  This
isn't without reservations, as some of the ambiguity present in the
string has been exploited to positive ends in the past, particularly by
browsers with less market share to avoid sites discriminating against
them.  On balance, losing that ambiguity and the associated bugs is a
net positive.

The main reservation we have regards the replication of information at
the HTTP layer into new header fields.  This is inefficient and provides
less accountability than a JS API.  The proposal has browsers send four
header fields in place of the one, which takes considerably more space.
Even with header compression, this adds non-trivial amounts of overhead
to every request.  Sites that request additional information with
Accept-CH will further inflate that.

The advantages we might realize from that design depend on the
structured information being more usable and reliable.  On the usability
side, more granular information can make caching strategies more
capable.  On the reliability side, use of structured fields means that
there should be fewer parsing irregularities and quirks.  This is a
general claim about Client Hints as a whole, just applied to User-Agent.

However, in the case of User-Agent, where the information is already
present, the trade-off is a little tougher.  Sites, and even caches,
have a means to access the information today.  It's inconvenient, but
various proprietary approaches exist to manage granular caching based on
User-Agent or Cookie; it might be worth exploring standardization of
other approaches to the same problem, if caching happens to be a major
motivating factor for the decomposition of the field.

The proponents are going to look at making the specification clearer
about its intent regarding new information.  With that understanding,
this change adjusts the position to reflect a position that is - for the
entire proposal - conflicted.

/cc @miketaylr @yoavweiss


hsivonen (changes requested):
I agree with the issue description, but I'd like the detail in activities.json to elaborate more on concerns related to the three new sent-by-default headers. I suggest:

UA Client Hints proposes that information derived from the User-Agent header field could be sent as new more structured headers only to HTTPS servers that opt into receiving that information, specifically to reduce the number of parties that can passively fingerprint users using that information. However, three of the new headers are sent unconditionally in every HTTPS request without explicit opt-in. We would prefer to progressively freeze the value of User-Agent HTTP header without replacement. We acknowledge the performance trade-offs this might introduce, but note that the performance characteristics of the proposed design are unclear and optimized almost exclusively for the very first connection to a site. The overheads involved add further uncertainty about performance as three new header fields are sent in every HTTPS request, plus fields for any information a site requests using the Accept-CH mechanism. Notably, the three added headers make it more likely some other new header that we might want to introduce ends up being the one that is the one too many in the light of server-side limits. Moreover, the benefit of edge caches being able to use the Vary mechanism on the new headers instead of varying on substrings of the User-Agent header isn't realized unless all clients send these new headers. Since edge caches are going to need a way to vary on the substrings of the User-Agent header for the foreseeable future, it would be more fruitful to standardize how the origin server expresses the User-Agent-based varying to the edge cache instead of expecting all clients to send the new headers. We conclude that using the proposed NavigatorUAData interface JS API is preferrable to use of header fields as it would better allow for auditing of the use of the information.

hsivonen (approved):
Thanks!

annevk (approved):
Should we also mention that Vary does not work with structured fields either so edge caching benefits only apply in some scenarios? Okay with me either way.

(The typo I fixed was preferrable -> preferable.)

martinthomson:
As we're already into walloftext territory with this, I think we're good to go.  That is, while I agree Vary is likely a poor fit even with these changes, but we don't need to enumerate all the failings of the system.

Issue #283, opened 2020-02-27
Event Timing


npm1:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Event Timing
* Specification or proposal URL: https://github.com/WICG/event-timing
* Mozillians who can provide input (optional): @mstange attended some of the presentations to the WebPerf WG and seemed to agree on the usefulness of the API. @bdekoz is currently attending WebPerf WG meetings so he may have thoughts.

### Other information

First Input Timing (event timing restricted to the first input event) already ships in Chrome. Chrome would like to ship full Event Timing sometime in the near future.

dbaron:
w3ctag/design-reviews#324 also has some feedback on this spec.

dbaron:
My personal reaction is that providing measurements on event processing times is a worthwhile problem to be solving.  The major concerns I can think of are:
* how well this addresses the use case of detecting event processing delay, and
* whether there are any security issues.

As to the first one, I think it's clear this is a metric that could be gamed by a hostile developer who's told to improve this metric but doesn't actually care about improving user-perceived performance, by deferring the work of handling the event until after the next paint.  I don't think that is, in itself, a problem.  However, people might have *other* reasons to defer work such that it ends up being after the next paint, and this might both (a) lead to the metric not measuring what was expected or (b) lead to things like "oh, we have to revert that change because it made our numbers on event timing much worse" when what really happened is that it started measuring the real work that happened.  I'm curious if we have any way to know how common that sort of deferral of work is.

That said, I would lean towards saying this is `worth prototyping` but I'd like to get input from others before marking it as such.

npm1:
Your concern is totally valid and it is indeed common to have the event handling work be asynchronous. This API only captures the synchronous work, and we do have plans to enable a developer to annotate their work and capture the async work (but that may happen via the User Timing API). Note that even if the work is synchronous, a developer can still get some benefits from the API by looking at the input delay (hardware timestamp to when input begins being processed). This part cannot be gamed easily and must be fast in order to provide a good interactivity experience (it's necessary, though of course not sufficient). A developer should look at high (80-99) percentiles of input delay and make sure that they are low. Then the developer can focus on making actual processing time fast, via this API if they know work is synchronous or in conjunction with User Timing if it's not.

npm1:
Update: we intend to augment this API with interactionID.

Explainer: https://gist.github.com/npm1/9c2b95ece116d9bcb4bc224155e23777
[...]

PR #589, opened 2021-10-28
Add Event Timing API as worth prototyping

sefeng211:
Position issue: #283 
 
We had an implementation already, so I think we should resolve it? 

@annevk

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -437,6 +437,18 @@
   },
   {
     "ciuName": null,
+    "description": "This specification defines APIs for observing latency of certain events triggered by user interaction.",
+    "id": "event-timing-api",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1667836",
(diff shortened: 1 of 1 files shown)


annevk (approved):
Thanks!

Hopefully this will be standardized soon. We actually shouldn't have shipped this in its current state: https://wiki.mozilla.org/ExposureGuidelines#Is_the_feature_covered_by_a_specification.3F.

npm1:
> Thanks!
> 
> Hopefully this will be standardized soon. We actually shouldn't have shipped this in its current state: https://wiki.mozilla.org/ExposureGuidelines#Is_the_feature_covered_by_a_specification.3F.

What do you mean by this? My two guesses are a) integrate it properly with DOM or b) move it into a proper WG, i.e. outside of WICG?

annevk:
@npm1 mainly the latter. (The former would hopefully be a consequence of that.)

Issue #194, opened 2019-07-29
Scroll To Text Fragment


nickburris:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Scroll To Text Fragment
* Specification or proposal URL: https://github.com/WICG/ScrollToTextFragment

### Other information
https://github.com/w3ctag/design-reviews/issues/392

annevk:
I guess it would help to summarize some of the issues I've seen:

1. Unclear processing model. It seems like this can be done without modifying the URL parser, but what exactly the processing model should be is not clear.
2. There are some risks with letting an attacker scroll a user to a chosen position and it's not clear the proposal goes to the fullest extent to eliminate these, e.g., https://github.com/w3ctag/design-reviews/issues/392#issuecomment-543148916.
3. The bits of the processing model I do understand mean that this needs to be stripped from the URL at some point to preserve user privacy (though the website can still look at the scroll offset to determine things). This creates a risk for non-implementing user agents and makes the new syntax less useful as a general way of extending fragments.
4. If this effectively triggers `window.find()` it would have been nice to clean up that first so they can use a shared primitive: https://github.com/whatwg/html/issues/3539.

(Also, if that highlight API comes of the ground I suppose it better be isolated from these highlights somehow.)

Overall though it seems like a useful feature to have.

domenic:
> Unclear processing model. It seems like this can be done without modifying the URL parser, but what exactly the processing model should be is not clear.

FWIW in my review of the spec I found the processing model fairly clear and precise. The entry point that you might find most clarifying are:

- https://wicg.github.io/ScrollToTextFragment/#processing-the-fragment-directive
- https://wicg.github.io/ScrollToTextFragment/#navigating-to-text-fragment

That said, it does appear that the draft spec does currently modify the URL parser/serializer. But the way in which it does so seems unobservable to any part of the ecosystem that doesn't specifically look at the "URL's fragment directive" concept. So I think it could be refactored transparently (i.e. with no testable effects) to put that parsing entirely in HTML (in particular in the first link above).

fred-wang:
>     4. If this effectively triggers `window.find()` it would have been nice to clean up that first so they can use a shared primitive: [whatwg/html#3539](https://github.com/whatwg/html/issues/3539).

I agree. I opened https://github.com/WICG/ScrollToTextFragment/issues/66

bokand:
Update:

[...]

PR #611, opened 2022-02-07
Update Text Fragments position

annevk:


tomayac:
As the implementer of the "official made by Google" [Firefox Link to Text Fragment extension](https://addons.mozilla.org/en-US/firefox/addon/link-to-text-fragment/), I can't wait for the browser to add proper support. Running the matching logic in user land is suboptimal, so more than happy for the browser to take over.

bholley (approved):

bholley:
And for those following along, @annevk recently wrapped up a thorough privacy analysis of this proposal, in consultation with various folks who have been researching its implications. The conclusion was generally positive, resolving the concerns we had at [1] and resulting in the updated position here.

[1] https://github.com/mozilla/standards-positions/issues/194#issuecomment-782740588

PR #652, opened 2022-06-16
Replace UTF quote with \u2019 per "./activities.py sort"

dholbert:
If you run "./activities.py sort" on the repo before this patch (to "Validate activities.json and write it out again in the canonical sorted order"), you get one change, which is precisely the change I'm making here: replacing a fancy UTF single-quote character with its encoded version, `\u2019`.

It seems we have several other such `\u2019` tokens already in this file, so I assume that representation is acceptable. So, this patch makes this replacement, so that `./activities.py sort` becomes a no-op.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1488,7 +1488,7 @@
     "mdnUrl": null,
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1690255",
     "mozPosition": "worth prototyping",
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

martinthomson:
Ugh, contractions...  Thanks for fixing, I'll[^1] take this and then we can consider whether c14n[^2] is worth enforcing.

[^1]: Hah! See what I did there!
[^2]: Again!

Issue #118, opened 2018-12-05
Container Queries


dbaron:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Container Queries
* Specification or proposal URL: There isn't really a proposal yet; there's just been a set of discussion, e.g., in https://github.com/WICG/container-queries
* Mozillians who can provide input (optional): @dbaron @jensimmons @tantek 

(There's a bunch more I intend to write about this... but not right now.)

marcoscaceres:
HI @dbaron @jensimmons @tantek.  Over at the WICG, we are wondering what we should do about the container queries incubation work. It hasn't moved in a while, and we are a bit unsure how to move it forward (though dependencies and similar technologies continue to progress). Folks that are closer to layout and css in general, I'm wondering if we could reach some kind of position on this.

bholley:
CC @heycam

dbaron:
Briefly, I think this in demand, useful, and is something worth working on.  But figuring out how to do it requires some serious work from somebody with solid knowledge of how browsers implement CSS, etc.

I think the eventual performance characteristics of a solution I'd hope to see aren't great (since it would require separating style+layout work for one part of the tree from that of another part), but they could still be substantially better than a Javascript-based solution built on top of today's technology (which would *duplicate* the work).

Tiedye:
Any update on this?

dholbert:
Sorry for the delayed response.

I agree with dbaron's earlier assessment above ("this is in demand, useful, and is something worth working on").

Now that we've got a relatively stable spec, we should definitely call this worth-prototyping.  And in fact we are working on prototyping it, under metabug https://bugzilla.mozilla.org/show_bug.cgi?id=1744221

bgrins:
Resolved by https://github.com/mozilla/standards-positions/pull/654

PR #654, opened 2022-06-16
Add Container Queries (worth prototyping)

dholbert:


dholbert:
Addresses #118.

bgrins (approved):
Thanks @dholbert, I think this text is good

PR #680, opened 2022-08-29
Positive, neutral, and negative positions

### LINKED ISSUE #242: Consider renaming worth prototyping (or adding a label)
As noted in #184 by @padenot and elsewhere by others, "worth prototyping" is not a great fit for things we're supportive of and already ship. "important" is the only alternative, but generally reserved for more important things.

Shall we rename "worth prototyping" to "supportive" or some such?

martinthomson:
A revision of the position taxonomy to use positive, neutral, and negative.

Closes #242.


ekr:
What's your proposed plan for dealing with our previous positions? I feel like we either need to go back and re-label or you need some reference for the old labels.

martinthomson:
This retroactively applies the new labels to old positions.  In reviewing those, I didn't find any that needed changes to make sense.

After merginf this, we will need to rename labels, which isn't perfect with respect to how some of GitHub's systems manage that.  But aside from those few corner cases, I don't anticipate any problems with doing that.

ekr:
So just to confirm, the plan is actually to relabel the old issues? I don't have a problem with that, I just want to make sure we don't have things still publicly saying (for instance) non-harmful but no definition for that.

README.md
```diff
@@ -13,20 +13,33 @@ elsewhere too.
...
-The currently possible positions are:
+We will seek to apply one of three labels to new specifications:
+
+- `positive` - Mozilla regards this work as a potential improvement to the web.
+- `neutral` - Mozilla is not convinced of the merits of this work, but does not see any significant negative potential.
+- `negative` - Mozilla believes that pursuing this work in its current form would not be good for the web.
+
+These positions will be based on our best assessment of the current state of the specification and
+the potential for it to benefit our users and the web as a whole.  We consider the intended purpose
+and the design fundamentals of each specification.  Incomplete documentation, minor issues, or lack
```
bgrins, line 28: Should be "lack of"?

bgrins:
> So just to confirm, the plan is actually to relabel the old issues?

Yes, confirming that's my understanding of the plan. And the PR doing that with the web UI - it barely touches index.html but the actual content is client side rendered from activities.json which has been rewritten to the new positions. As for issues which are directly labeled in github and not part of the web UI, those would be updated automatically after renaming as per https://github.com/mozilla/standards-positions/pull/680#issuecomment-1229621582.

README.md
```diff
@@ -13,20 +13,33 @@ elsewhere too.
...
+interoperability testing are not reasons for a negative position, provided that intent is clear.
+
+We might also use the following labels for specifications where we do not have enough information or
```
bgrins, line 31: I like splitting these out into a group separate from the others. But the old definition for `defer` includes a lot of discretion for the reason to not set a position - it's just "does not intend to" (and even stronger with "at all"). This new group header text narrows this quite a bit by basing it on missing information. I can think of cases where any information we'd need to form a position is accessible to us, but we choose not to invest the time to do it. In some sense the text here is still accurate - we don't have enough information, but that's because we aren't seeking it. And I think defined this way it may be construed as a desire for people looking for our position to just send over more information if something is deferred.

WDYT? If we want to keep closer semantics to what we have today we could prefix the group with something simpler like "We may use the following labels for specifications where we haven't formed a position".

activities.json
```diff
@@ -317,7 +317,7 @@
   },
   {
     "ciuName": null,
-    "description": "This proposal adds an API for prompting and querying the user\u2019s contacts for one or more items with a handful of contact properties.",
+    "description": "This proposal adds an API for prompting and querying the user’s contacts for one or more items with a handful of contact properties.",
```
bgrins, line 320: IIRC this unfortunately breaks the activities.py tool (as per https://github.com/mozilla/standards-positions/pull/652)

activities.json
```diff
@@ -733,10 +733,10 @@
   },
   {
     "ciuName": "mdn-api_keyboardlayoutmap",
-    "description": "This specification defines an API that allows websites to convert from a given code value to a valid key value that can be shown to the user to identify the given key. The conversion from code to key is based on the user\u2019s currently selected keyboard layout. It is intended to be used by web applications that want to treat the keyboard as a set of buttons and need to describe those buttons to the user.",
+    "description": "This specification defines an API that allows websites to convert from a given code value to a valid key value that can be shown to the user to identify the given key. The conversion from code to key is based on the user’s currently selected keyboard layout. It is intended to be used by web applications that want to treat the keyboard as a set of buttons and need to describe those buttons to the user.",
```
bgrins, line 736: Ditto

activities.json
```diff
@@ -1018,11 +1018,11 @@
   },
   {
     "ciuName": null,
-    "description": "This document specifies modifications to Fetch which are intended to mitigate the risks associated with unintentional exposure of devices and servers on a client\u2019s internal network to the web at large.",
+    "description": "This document specifies modifications to Fetch which are intended to mitigate the risks associated with unintentional exposure of devices and servers on a client’s internal network to the web at large.",
```
bgrins, line 1021: Ditto - last one :)

README.md
martinthomson:
```suggestion
and the design fundamentals of each specification.  Incomplete documentation, minor issues, or lack of
```

activities.json
martinthomson: Thanks for catching these.  `./activities.py sort` fixed them all up.

README.md
martinthomson: Yeah, I think that is easier.  I wanted to give more context, but it turns out that trying to enumerate the list failed to be complete.  Sometimes we just don't want to, but putting that into the README doesn't sound good.

bgrins (approved):
LGTM

tantek (approved):
Looks good to me with the updated edits.

PR #696, opened 2022-10-12
Point trusted types link to new address

KyFaSt:


tantek (approved):
This seems ok to me, however, shouldn't https://w3c.github.io/webappsec-trusted-types redirect automatically?

bgrins:
Good catch - looks like https://github.com/w3c/webappsec-trusted-types redirects now to https://github.com/w3c/trusted-types so makes sense the page url changed as well.

PR #719, opened 2022-12-13
Update the details of WebTransport entry

omasanori:
This PR updates the WebTransport Protocol Framework entry to refer to the latest information.

I am sorry if this is not a preferred way to amend a published position entry.


martinthomson (approved):
Thanks for the updates.  Much appreciated.

PR #733, opened 2023-01-18
use LS/ED URLs

tantek:
add tip for Mozilla folks, spec URLs in activities.json PRs should use living standard / editor's draft URLs


martinthomson (approved):
Yeah, not a lot of good options for recording stuff like this.  We could update the issue template, but that seems unnecessary (not a lot of /TR/ links hitting that path).

PR #742, opened 2023-01-31
Add Explainer URL to issue template

### LINKED ISSUE #739: Meta: add issue template fields/templates to explain the maturity of the spec
In discussing with @tantek, it seems like you want Chrome to file standards positions around the time we have IDL and an API shape (which we usually do inside the explainer, roughly during [step 2 of the Blink process](https://www.chromium.org/blink/launching-features/#step-2-prototyping)), which is often before we write a specification (roughly the [end of step 4 or early in step 5](https://www.chromium.org/blink/launching-features/#step-4-widen-review)). If that's accurate, this repo's [issue template (or templates)](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates) should have a field where we can say whether the URL refers to an explainer or full specification, and probably the rough timeline we expect before we'd want to ship something.

zcorpan:
Fixes #739.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -2,7 +2,8 @@
 ## Request for Mozilla Position on an Emerging Web Specification
 
 * Specification Title: 
-* Specification or proposal URL: 
+* Specification or proposal URL (if available): 
+* Explainer URL (if available): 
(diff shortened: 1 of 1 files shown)


tantek (approved):
LGTM

PR #748, opened 2023-02-15
create entry for Autoplay Policy Detection

alastor0325:


zcorpan (approved):

tantek (approved):
LGTM, accurately reflects consensus in issue 675. I think due to the fingerprinting concerns raised it's worth an explicit dashboard entry.

PR #786, opened 2023-04-24
I see a red door

martinthomson:
A repeatable process for making code less variable:

```
python -m venv .venv
. .venv/bin/activate
pip install black
python -m black .
```


zcorpan (approved):

zcorpan:

PR #787, opened 2023-04-24
Make the better URL feature work properly 

martinthomson:
This uses `Exception.args` rather than trying to index the `Exception` instance directly.


zcorpan:
This includes the commit in https://github.com/mozilla/standards-positions/pull/786 so will merge after that

zcorpan:

PR #793, opened 2023-04-29
chore: port GitHub issue form from Webkit

nschonni:
The WebKit repo moved over to the the new issue template. I took a quick pass at porting their template, but feel free to push any more changes

You can see the rough render by looking at the file https://github.com/mozilla/standards-positions/blob/8a56a4958769ddbbd0c6a36c440319871322526f/.github/ISSUE_TEMPLATE/request-for-position.yml


zcorpan:
@bgrins are there any scripts or so that would need updates for this change?

bgrins:
The only thing I can think of is [this script](https://github.com/bgrins/compat-utils/blob/3de6a28ced6b6df83913ca198b9e7d377ae38284/write-mozilla-positions.js#L105) which was used to seed a project and to keep a combined position file (combining activities.json and labeled issues). If we want that CSV to stay updated we'd need to change the script to look for both the old and new text for each field but I don't mind doing that.

.github/ISSUE_TEMPLATE/request-for-position.yml
```diff
@@ -0,0 +1,84 @@
...
+description: If you're working on a spec and would like to know the Mozilla's position on it, this is the form for you!
+body:
+  - type: markdown
+    attributes:
+      value: "## Mozillians who can provide input"
+  - type: input
+    id: mozillians
+    attributes:
+      label: "Mozillians"
+      description: "Who from the WebKit community do you think is best positioned to review this spec? It's okay if you don't know."
```
bgrins, line 11: WebKit -> Mozilla

tantek (changes requested):
s/@dbaron//

tantek (changes requested):
In reviewing the changes to CONTRIBUTING.md and the deletion of ISSUE_TEMPLATE.md in particular, this change is a UX regression (how folks, especially developers filing issues, interact with standards-positions) in a couple of ways that I'd like to see fixed before landing this:

1. The custom link for "open a new issue" in CONTRIBUTING.md belies the fact that it's perfectly reasonable (more likely) for someone to come to the repo, click Issues, and click New issue, than be burdened with hunting through CONTRIBUTING.md every time for one specific text link embedded in prose. Expected behavior: using GitHub UI to click "New issue" should "just work". 
2. Currently developers can post a new issue via the GitHub API as noted in CONTRIBUTING.md. I’ve done this myself and it works great. Removing that is a regression.

Thus the requested fixes are:
1. Restore ISSUE_TEMPLATE.md
2. Update it statically with the fields from the new issue form
3. Restore the text/links in CONTRIBUTING.md regarding "(If posting via GitHub API, please copy the template and fill it out)." We should continue to support this use-case.

nschonni:
You don't need to use the special link, when clicking on new, you get the template chooser as seen on the WebKit repo
image

zcorpan:
This would now require further changes to the new gh-data.py script to also support the new template. I think we should leave this as-is for now.

PR #816, opened 2023-06-08
Dark mode dashboard

### LINKED ISSUE #814: Dark mode on your website
this is not a request for Mozilla Position on an Emerging Web Specification
but a request for https://mozilla.github.io/standards-positions/ to support dark mode.

The site is too bright at night.

and hopefully also make use of `color-scheme` with either of this solutions:

`<meta name="color-scheme" content="dark light">`
```css
@media (prefers-color-scheme: dark) {
  :root {
    color-scheme: dark;
  }
}
```

- https://web.dev/color-scheme/

martinthomson:
This isn't so trivial, but likely only because I don't know how to drive bootstrap.  This just overrides the rules for stuff we use.

Feedback on color choice welcome; I just picked something that looked OK on my screen.

Closes #814.

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,9 +30,8 @@
         padding-bottom: 5px;
       }
 
-      table.dataTable > tbody > tr:target,
-      table.dataTable > tbody > tr.shown:target + tr {
-        background: #fd0;
+      table.dataTable > tbody > tr:is(:target, .shown:target + tr) {
+        background: #fdeb8b;
       }
 
       tr:not(:hover):not(:focus-within) > .link-control > a {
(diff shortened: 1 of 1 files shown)


martinthomson:
Example:
image

zcorpan:
This looks OK I think (but I'm not a designer). Maybe `mark` should have different colors?

I was also going to suggest to check for color contrast issues (with Accessibility Inspector), but light mode already has some issues there so could be a follow-up.

bgrins:
Looks like if we were using bootstrap >= 5.3 there's some built in support for dark theme and custom variables that could be set: https://getbootstrap.com/docs/5.3/customize/color-modes/. Upgrading to a new version may be more work than this patch though

zcorpan:

PR #868, opened 2023-08-16
Global Privacy Control (positive)

bvandersloot-mozilla:
There are two ciu links that are relevant... we could just pick one, but I left it null.

This should probably have @martinthomson's eyes, particularly on the mozPositionDetail.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -688,6 +688,18 @@
     "url": "https://wicg.github.io/get-installed-related-apps/spec"
   },
   {
+    "ciuName": null,
+    "description": "This spec defines a signal that conveys a person's request to websites and services to not sell or share their personal information with third parties.",
+    "id": "gpc",
+    "mozBugUrl": null,
+    "mozPosition": "positive",
+    "mozPositionDetail": "This signal defined in this specification provides users with a way to opt-out of the disclosure of their information to third parties once per profile in a way that is legally enforced in some jurisdictions, and is being considered in future regulations. This is a win for user privacy now that there are regulatory teeth coming in.",
+    "mozPositionIssue": 867,
+    "org": "W3C",
+    "title": "Global Privacy Control (GPC)",
+    "url": "https://privacycg.github.io/gpc-spec/"
+  },
+  {
     "ciuName": "imports",
     "description": "HTML Imports are a way to include and reuse HTML documents in other HTML documents.",
     "id": "html-imports",
--- a/activities.json
+++ b/activities.json
@@ -691,7 +691,7 @@
     "ciuName": null,
     "description": "This spec defines a signal that conveys a person's request to websites and services to not sell or share their personal information with third parties.",
     "id": "gpc",
-    "mozBugUrl": null,
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1848951",
     "mozPosition": "positive",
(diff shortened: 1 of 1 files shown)


tantek (changes requested):
This looks good overall. mozBugUrl should probably be https://bugzilla.mozilla.org/show_bug.cgi?id=1848951 tho (unless there's a more appropriate bug)

tantek (changes requested):
Would also be good to add an explicit
"mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/API/Navigator/globalPrivacyControl"
assuming that's the right page

bvandersloot-mozilla:
> Would also be good to add an explicit "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/API/Navigator/globalPrivacyControl" assuming that's the right page

There is also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-GPC. This is the same issue as the CanIUse links. Both are valid signals so I don't really want to link to one and not the other.

bvandersloot-mozilla:
I just realized that arrays are accepted in the `ciuName` and `mdnUrl` fields. Including both there.

However, this leads to a styling issue because we have more than 4 icons (mdnUrl with an array produces multiple icons) so we cause a text wrap in that table row.

martinthomson (approved):
The MDN URL is probably worth adding, even though it is just one of two.  Either is fine, they each contain content for the other.

tantek (changes requested):
For the mdnUrl, better to link to just the one higher level page, https://developer.mozilla.org/en-US/docs/Web/API/Navigator/globalPrivacyControl in this case, which itself links to the specific Sec-GPC header page.

tantek (approved):

tantek (approved):
LGTM. The only remaining oddness is use of "W3C" as the org for something which is not at a w3.org URL (rather than a github.io URL) and is not standards track yet (in a CG only). We can allow this to land like this in the dashboard since we are already actively pushing for GPC to be included in the Privacy WG charter (openly in development with GPC included) which would make it standards track. When Privacy WG starts up and publishes a FPWD of GPC, we should update this entry accordingly. Can you track that and update this entry accordingly when the time comes @bvandersloot-mozilla? Thanks.

activities.json
```diff
@@ -687,6 +687,19 @@
...
+  {
+    "ciuName": null,
+    "description": "This spec defines a signal that conveys a person's request to websites and services to not sell or share their personal information with third parties.",
+    "id": "gpc",
+    "mdnUrl": ["https://developer.mozilla.org/en-US/docs/Web/API/Navigator/globalPrivacyControl"],
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1848951",
+    "mozPosition": "positive",
+    "mozPositionDetail": "This signal defined in this specification provides users with a way to opt-out of the disclosure of their information to third parties once per profile in a way that is legally enforced in some jurisdictions, and is being considered in future regulations.",
+    "mozPositionIssue": 867,
+    "org": "W3C",
```
martinthomson, line 699: I missed this, but this really needs to be a proposal.  We need to be consistent.  If this changes (and it might), then we can update.

```suggestion
    "org": "Proposal",
```

bvandersloot-mozilla:
Thanks for the catch- the subtleties of the org field's semantics were not clear before, but are now. I'll follow it and update the dashboard if/when there is a FPWD in a W3C WG.

tantek (approved):
Thanks for the details. Appreciated. I don't see anything else to fix. Will leave it to @martinthomson to merge.

PR #894, opened 2023-09-28
Service Worker Static Routing API (positive)

### LINKED ISSUE #828: ServiceWorker static routing API

* Specification Title: ServiceWorker static routing API
* Specification or proposal URL (if available): 
* Explainer URL (if available): https://github.com/yoshisatoyanagisawa/service-worker-static-routing-api

* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 

### Other information

This API allows developers to configure the routing, and allows them to offload simple things ServiceWorkers do. If the condition matches, the navigation happens without starting ServiceWorkers or executing JavaScript, which allows web pages to avoid performance penalties due to ServiceWorker interceptions.

Starting ServiceWorkers is known to be a slow process, and web users need to wait for its startup if the ServiceWorker intercepts loading the page resources. At the same time, the ServiceWorker brings flexibility to the transport layer, and it behaves as a client-side proxy. Developers can implement offline support or provide client-side content modification with it. Currently, ServiceWorkers intercept all requests for pages within their scope, which brings a performance penalty within the scope.

asutherland:
Closes #828

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1319,6 +1319,19 @@
     "url": "https://tools.ietf.org/html/draft-ietf-dnsop-svcb-httpssvc"
   },
   {
+    "ciuName": null,
+    "description": "The initial proposal allows ServiceWorkers to define routes using URLPattern to bypass the ServiceWorker, avoiding spurious ServiceWorker wake-ups and avoiding the additional latency of sending requests to the ServiceWorker that it will not handle.  Longer term, the mechanism could also support consulting the Cache API without needing to wake up a ServiceWorker.",
+    "id": "service-worker-static-routing-api",
+    "mdnUrl": null,
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1855580",
+    "mozPosition": "positive",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -1318,6 +1318,19 @@
...
+    "id": "service-worker-static-routing-api",
+    "mdnUrl": null,
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1855580",
+    "mozPosition": "positive",
+    "mozPositionDetail": "A longstanding performance concern for ServiceWorkers and the sites using them is that there is no way to skip going to a ServiceWorker for a controlled page when there is a \"fetch\" handler present.  The Static Routing API has been recognized as an ideal solution to address the problem versus other proposals like adding new attributes to HTML tags or arguments to the fetch API.  The static routing API had been discussed extensively in ServiceWorker WG meetings as a reasonable option but very complex to implement and specify, which is why it was not part of the original spec.  With the introduction of the URLPattern API, the Static Routing API is thankfully even easier to implement and specify.  The evolutionary approach where the Static Routing API will only start with a network source that bypasses the ServiceWorker is appropriately pragmatic.",
+    "mozPositionIssue": 828,
+    "org": "W3C",
```
asutherland, line 1330: I have marked this as W3C as I think this would be integrated into the ServiceWorker spec at the W3C ServiceWorkers WG, especially since at WG meetings we had discussed the "declarative routing API" rather extensively and it seemed certain we would do it but not as part of 1.0, as it were.  But I may be misunderstanding the correct use of this field.

asutherland: org has been changed to "Proposal"

asutherland:
> I'm ok with marking this as 'org' W3C only if the 'url' field points to a W3C WG spec or an issue on said spec.
> 
> If we want to keep the 'url' field as linking to a WICG issue, then the 'org' should be "Proposal".

I think the locus of activity is current on the WICG repo which has its own issues, so it probably makes sense to keep the URL as the WICG repo and mark this as Proposal, which I've now done.

tantek (approved):

Issue #840, opened 2023-06-30
Lazy loading for iframes


zcorpan:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Lazy loading for iframes
* Specification or proposal URL (if available): https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-loading
* Explainer URL (if available): https://github.com/scott-little/lazyload
* Caniuse.com URL (optional): https://caniuse.com/loading-lazy-attr (images and iframes)
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=1622090
* Mozillians who can provide input (optional): @hiikezoe @emilio

### Other information

Also see https://github.com/mozilla/standards-positions/issues/151

zcorpan:
We support lazy-loaded images, and the feature makes sense for `iframe` for similar reasons. Initial page load is faster and more lightweight if out-of-view iframes are lazy-loaded. It's possible for sites to implement lazy-loading with JS but a declarative solution is easier to use and allows browsers to control when to load things.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1622090#c10 for an example of real-world impact.

Suggested position: positive.

emilio:
I agree fwiw.

martinthomson:
@zcorpan did you intend to add this to the dashboard?

PR #911, opened 2023-10-19
Change lazy-loading-attr to also cover iframes

zcorpan:
See #840

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -888,7 +888,7 @@
     "ciuName": "loading-lazy-attr",
     "description": "Enabling images and iframes to be fetched at a later time, when the user is likely to look at them.",
     "id": "loading-lazy-attr",
-    "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading",
+    "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes",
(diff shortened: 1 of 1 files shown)


tantek (changes requested):
These changes look good.

I would additionally suggest updating the mdnUrl to https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes which covers both images and iframes

Ideally we'd link to both mozBugUrls (image and iframe lazy loading) but I'm not sure if multiple bug URLs are supported

activities.json
```diff
@@ -886,16 +886,16 @@
   },
   {
     "ciuName": "loading-lazy-attr",
-    "description": "Enabling images to be fetched at a later time, when the user is likely to look at them.",
+    "description": "Enabling images and iframes to be fetched at a later time, when the user is likely to look at them.",
     "id": "loading-lazy-attr",
     "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading",
```
zcorpan, line 891:
```suggestion
    "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes",
```

zcorpan:
@tantek changed mdnUrl. Yeah I think we can only point to one bug.

tantek (approved):
LGTM

PR #917, opened 2023-10-26
Submit Ecma TC39 proposal-decimal

jessealama:


zcorpan:
Thanks for contributing! For TC39 proposals, there's no need to use Mozilla standards-positions, see https://mozilla.github.io/standards-positions/#tc39-stage-3

For future requests for a standards position (not TC39), please [file an issue](https://github.com/mozilla/standards-positions/issues/new) first.

PR #936, opened 2023-12-11
Update standards position on Trusted Types - fixes #20

### LINKED ISSUE #20: Trusted Types

mozfreddyb:


activities.json
```diff
@@ -1573,8 +1573,8 @@
...
     "id": "trusted-types",
     "mozBugUrl": null,
-    "mozPosition": "neutral",
-    "mozPositionDetail": "The API could be used to harden sites against certain cross-site scripting issues, but it is sufficiently complex that we are concerned that it will not be suitable for many sites.",
+    "mozPosition": "positive",
+    "mozPositionDetail": "Mozilla believes that preventing DOM-based XSS is an important security goal. The track record of preventing DOM-based XSS is convincing. That being said, the Trusted Types specification is also providing API methods with little or unknown value and uptake, like getPropertyType, getAttributeType. Additionally, there are features in the Chrome implementations that are not yet standardized, like the beforepolicycreation event. These should ideally be properly standardized (based on a proven need) or deprecated and removed. Dealing with inscrutable third-party dependencies or external JavaScript has been a major concern of security and enforcing reasonable boundaries is a promising approach. We believe that runtime checking of untrusted HTML might pair well with a Sanitizer API that should be complementary.",
```
martinthomson, line 1577: The track record of *what* preventing DOM-based XSS?

This seems a little overwrought for a positive position.  It reads as somewhat negative even.

Can you break down the critique on the issue and provide a shorter summary here?

mozfreddyb: I shortened it. How about this?

> Mozilla believes that preventing DOM-based XSS is an important goal and the track record of Trusted Types preventing it is convincing. The specification will likely require some work towards a clearer API shape. Also, Chrome seems to have added feature that are not yet specified, like the beforepolicycreation event. These should ideally be properly standardized (based on a proven need) or deprecated and removed. We believe that runtime checking of untrusted HTML might pair well with a Sanitizer API that should be complementary.

martinthomson: Much better.  How about a little shorter yet:

```suggestion
    "mozPositionDetail": "Mozilla believes that preventing DOM-based XSS is an important security goal. The track record of preventing DOM-based XSS is convincing. Dealing with inscrutable third-party dependencies or external JavaScript has been a major concern of security and enforcing reasonable boundaries is a promising approach. We have some reservations about some features in the Chromium implementation, which need to be validated and standardized or removed.",
```

I struggled with the sentence about Sanitizer, but I don't think that we need to talk about Sanitizer here.  That's a comment that might go well on #20 though.

The same applies to:

> That being said, the Trusted Types specification is also providing API methods with little or unknown value and uptake, like getPropertyType, getAttributeType. Additionally, there are features in the Chrome implementations that are not yet standardized, like the beforepolicycreation event. These should be properly standardized (based on a proven need) or deprecated and removed.

Those details can go on the issue discussion (#20).

mozfreddyb: Agreed. I'll respond there!

martinthomson (approved):

PR #940, opened 2023-12-13
Update actions; cache pip

martinthomson:


tantek (approved):
Seems like a straightforward upgrade.

Issue #687, opened 2022-09-06
Re-request for postion on Web Serial/Serial API


NODeeJay:
Since https://github.com/mozilla/standards-positions/issues/336 overheated and got closed by @bgrins, also my re-request https://github.com/mozilla/standards-positions/issues/336#issuecomment-1173177328 on the position got closed.

Based on https://github.com/mozilla/standards-positions/blob/main/CONTRIBUTING.md I am hereby asking Mozilla / @martinthomson to update the position on this topic.

Why:
Many solutions have been provided in the meantime to fulfill this request in a secure and safe way, and also many ways of securing the access itself. Hiding these settings for the experienced users is a better way of implementation, than no implementation at all, for example:

- about:config
- yes I understand the risks.
- Activate WebSerial, WebUSB, Web....
- only activate for SSL sites
  - ignore invalid certificates or present a "I understand the risk page" when an invalid certificate is presented
- only activate in local networks
- only activate for whitelist IP/hostname
- activate for all but blacklist IP/hostname
- or generally the same list of exceptions like for "unsafe" pages
- limit to the following COM ports/port range (this prevents access to system internal devices at all) or
- exclude the following COM ports

Furthermore, I am pointing out your core values:
Our mission is to ensure the Internet is a global public resource, open and accessible to all. An Internet that truly puts people first, where individuals can shape their own experience and are empowered, safe and independent.
Most important here is that **you empower us in a safe and independent way**!

I therefore kindly urge you to revise your stance on this topic.

Thank you.

As this thread should evaluate possible solutions and their technical integration I urge anyone participating to follow https://en.wikipedia.org/wiki/Wikipedia:Rational_debate aside of the [code of conduct](https://github.com/mozilla/standards-positions/blob/871e866a7683d75f6458d23d5183468847a8fe2f/CODE_OF_CONDUCT.md) and [Mozilla guidelines ](https://github.com/mozilla/standards-positions/blob/871e866a7683d75f6458d23d5183468847a8fe2f/CONTRIBUTING.md) and keep this thread solution oriented. The worst would be another closure.

bholley:
Just a quick level-set on Mozilla's stance here.

Over the past year or so, we've been experimenting as time permits with various approaches to improve the safety of inherently-powerful APIs, in line with the [Web Vision](https://www.mozilla.org/en-US/about/webvision) document we published earlier this year. We've prototyped a few different designs for this with WebMIDI, and while that project is still in progress, it seems to be converging on something workable.

[...]

PR #959, opened 2024-01-10
Update Serial position

bgrins:


bgrins:
Fixes #687

tantek (approved):
I think this accurately reflects our updated experience and more cautious method for approaching these sorts of specifications. Approving without merging to get more reviews because this is a more long-running / sensitive shift than usual.

bholley:
LGTM

Ferdi265:
With this merged, will Mozilla reconsider their stance of not implementing WebSerial (as per [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=webserial)) and potentially implement it in a similar add-on gated way to WebMIDI?

bholley:
> With this merged, will Mozilla reconsider their stance of not implementing WebSerial (as per [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=webserial)) and potentially implement it in a similar add-on gated way to WebMIDI?

This PR represents a reconsideration of our stance. We'd take a patch, assuming we think the proposed language in the prompt is sufficiently clear and assuming the patch otherwise met all of our usual standards.

nyabinary:
> With this merged, will Mozilla reconsider their stance of not implementing WebSerial (as per [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=webserial)) and potentially implement it in a similar add-on gated way to WebMIDI?

The bug should be reopened though as tracking though at least, right?

PR #960, opened 2024-01-10
s/webrtc-insertable-streams/webrtc-encoded-transform/ fixes broken link

jan-ivar:
The old link was broken (and forwarding appears to have stopped).

The spec is now at https://www.w3.org/TR/webrtc-encoded-transform/ not to be confused with https://www.w3.org/TR/mediacapture-transform/ (tiny edit of description to clarify this is not the latter)

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -1848,16 +1848,16 @@
   },
   {
     "ciuName": null,
-    "description": "This API defines an API surface for manipulating the bits on MediaStreamTracks being sent via an RTCPeerConnection.",
-    "id": "webrtc-insertable-streams",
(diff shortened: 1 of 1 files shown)


tantek (approved):
@jan-ivar thanks for catching this. LGTM.

Approving & merging to get this fix into our dashboard, however can you work with your WG team contact to fix the old URL https://w3c.github.io/webrtc-insertable-streams/ to redirect properly instead of 404ing? I believe W3C has committed to keeping these URLs working so they should. Feel free to cc me if you think it will help.

PR #986, opened 2024-02-15
:has() selector (Positive)

### LINKED ISSUE #528: :has() pseudo class

* Specification Title: The Relational Pseudo-class: ':has()'
* Specification or proposal URL: https://www.w3.org/TR/selectors-4/#relational
* Caniuse.com URL (optional): https://caniuse.com/css-has
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=418039

### Other information

The :has() pseudo class is a selector that specifies elements which has at least one element that matches the relative selector passed as an argument.

Unlike other selectors, it gives a way to apply a style rule to preceding elements (preceding siblings / ancestors / preceding siblings of ancestors) of a certain elements.

This difference is attractive to web developers, but also it generates lots of concerns mainly about performance and complexity.

We thought that, clarifying those concerns would be helpful for the discussion. So we started to check feasibility on blink, and were able to get some meaningful and reasonable results from this step. Based on the results, we are going to move forward by prototyping.

You can get the details about the current status in here.
* Explainer: https://github.com/Igalia/explainers/tree/main/css/has
* Test summary: https://css-has.glitch.me/

dshin-moz:
Closes #528

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -313,7 +313,7 @@
     "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/CSS/:has",
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=has-pseudo",
     "mozPosition": "positive",
-    "mozPositionDetail": "We acknowledge the developer demand for this selector, and the implementation is shipping as of 121. However, we are still concerned about the worst-case performance of DOM tree mutation when using the selector.",
+    "mozPositionDetail": "We recognize that :has is something that a lot of people want. That power does come with performance costs, like the potential for very poor performance when there is DOM tree mutation. Overall, the utility of the selector justifies this risk, but we might need to do more to help developers avoid the worst problems.",
     "mozPositionIssue": 528,
     "org": "W3C",
     "title": "CSS Relational Pseudo-Class (:has())",
--- a/activities.json
+++ b/activities.json
@@ -319,6 +319,19 @@
     "url": "https://drafts.csswg.org/css-cascade-6/#scoped-styles"
   },
   {
+    "ciuName": "css-has-selector",
+    "description": ":has selector lets authors select elements that anchor at least one element that matches its inner relative selector.",
+    "id": "css-has-selector",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -306,6 +306,19 @@
...
     "url": "https://drafts.css-houdini.org/css-properties-values-api-1#at-property-rule"
   },
+  {
+    "ciuName": "css-has-selector",
+    "description": ":has selector lets authors select elements that anchor at least one element that matches its inner relative selector.",
+    "id": "css-has-selector",
+    "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/CSS/:has",
+    "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=has-pseudo",
+    "mozPosition": "positive",
+    "mozPositionDetail": "We acknowledge the developer demand for this selector, and the implementation is shipping as of 121. However, we are still concerned about the worst-case performance of DOM tree mutation when using the selector.",
```
zcorpan, line 316: I think we generally try to not reference Firefox shipping status in standards-positions rationale. Try to answer if the feature is good for the web.

Also, I think this comes across as basically negative. "We had to implement it, but we don't like it."

zcorpan (changes requested):

tantek (approved):
LGTM, verified this works

tantek (approved):
LGTM with the patch updates here.

activities.json
martinthomson: How about:

```suggestion
    "mozPositionDetail": "We recognize that :has is something that a lot of people want. That power does come with performance costs, like the potential the worst-case performance resulting from DOM tree mutation to be severe. Overall, the utility of the selector justifies this risk, but we might need to do more to help developers avoid the worst problems.",
```

tantek (approved):
LGTM with @martinthomson's suggested change.

zcorpan (approved):

czarnajusta23-boop (commented):
activities.json6

Issue #775, opened 2023-04-13
Request for Mozilla Position: Zstandard compression format for Content-Encoding


nidhijaju:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Zstd Content-Encoding
* Specification or proposal URL (if available): https://datatracker.ietf.org/doc/html/rfc8878
* Explainer URL (if available): https://docs.google.com/document/d/1aDyUw4mAzRdLyZyXpVgWvO-eLpc4ERz7I_7VDIPo9Hc/edit?usp=sharing
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): https://bugzilla.mozilla.org/show_bug.cgi?id=zstd
* Mozillians who can provide input (optional): @martinthomson 

### Other information
[Zstandard](https://facebook.github.io/zstd/), or "zstd", is a data compression mechanism, described in [RFC8878](https://datatracker.ietf.org/doc/html/rfc8878), providing high compression ratios. The proposal is to add "zstd" to the tokens in the `Accept-Encoding` header, and decompress with Zstandard when the token is seen in the `Content-Encoding` header. It is anticipated to make pages load faster and reduce serving costs, by reducing CPU and/or bandwidth consumption, especially for dynamic content. 

Chrome is currently evaluating implementing this, and is interested in other browsers' position on the specification.

ricea:
I just found out that this was previously closed as `defer`: https://github.com/mozilla/standards-positions/issues/105

I do not have a lot of new information to contribute, except that zstd adoption has continued to grow.

rektide:
I noticed a couple items down is #160, which is for adding Compression Dictionary support to websites. This is WIP, but it seemingly could/would overcome a primary cause for the `defer` in #105 / here.

I'll comment in #105 as well.

martinthomson:
We can skip the dashboard for this one.

jesup:
FYI, we just landed zstd support in Nightly

martinthomson:
A timely reminder that we probably want to update the dashboard.  Note that this and that are subtly different things, so we might need to tweak a few things.

PR #1042, opened 2024-06-18
Switch position on zstd to positive

jesup:
Feel free to bikeshed on the wording.  Switched the issue to 775 where we agreed to ship zstd


tantek (approved):
Updated summary and position reflect latest discussions in #775.

Only nit I have is that https://datatracker.ietf.org/doc/html/rfc8478 says "Category: Informational" so should this be labeled a "Proposal" rather than "IETF" since as the RFC says:

> Despite use of the word "standard" as part of its name, readers are
   advised that this document is not an Internet Standards Track
   specification; it is being published for informational purposes only.

And if it's actually (being) interoperably implemented, who is taking care of moving this to a standards-track working group to get it officially (not just de facto) standardized?

adamroach:
[As the person who wrote the quoted text](https://datatracker.ietf.org/doc/draft-kucherawy-dispatch-zstd/history/): based on the history of this document in the IETF, I'm pretty comfortable asserting that no one is likely to ever have the appetite to push the zstd specification onto a standards track. Unless some other SDO takes up the work of publishing a standard, its status as "informational" is likely permanent.

I agree with Tantek that "proposal" is the correct category for this feature, per my understanding of feature characterizations.

zcorpan (approved):

PR #1063, opened 2024-08-28
Switch to YAML and GitHub data

zcorpan:
This changes the data format for activities from JSON to YAML, and uses information from the GitHub issues (the OP and labels). The data is presented on a new dashboard website. There's a workflow to push the `merged-data.json` file to `gh-pages` to publish.


gh-data.py
```diff
@@ -0,0 +1,159 @@
...
+def process_labels(labels):
+    position = None
+    venues = []
+    concerns = []
+    topics = []
+
+    for label in labels:
+        # Position
+        if label["name"] == "blocked":
+            assert position is None
```
martinthomson, line 50: It might pay to turn these assertions into warning messages instead.  We can have a `-Werror` analogue so that we can ensure that they surface, but the idea that this whole process aborts if we mess up the labeling worries me.

gh-data.py
```diff
@@ -0,0 +1,159 @@
...
+            position = "blocked"
+        elif label["name"].startswith("position: "):
+            assert position is None
+            position = label["name"][len("position: ") :]
```
martinthomson, line 54:
```suggestion
            position = label["name"].removeprefix("position: ")
```

&c...

Though maybe `label["name"].split(": ")` is a better implement.

gh-data.py
```diff
@@ -0,0 +1,159 @@
...
+        data = []
+        page = 1
+        while True:
+            try:
+                response = requests.get(
+                    f"https://api.github.com/repos/mozilla/standards-positions/issues?direction=asc&state=all&per_page=100&page={page}",
+                    timeout=5,
+                )
+                response.raise_for_status()
+            except Exception:
```
martinthomson, line 136: It might pay to check for rate limits.  We have a fair number of issues and I have first-hand experience with hitting those limits.  Look at [this code](https://github.com/MikeBishop/archive-repo/blob/2683167304c4a14f11ee370b535fc23620044a6a/archive-repo/archive.py#L459-L482) for some clues on how to manage that.

Also, this request is not using `$GITHUB_TOKEN`, but I've found that doing that helps a fair bit.  It means setting a [header](https://github.com/MikeBishop/archive-repo/blob/2683167304c4a14f11ee370b535fc23620044a6a/archive-repo/archive.py#L730-L733), but it means that you are less likely to hit rate limits.

activities.yml
```diff
@@ -0,0 +1,2798 @@
+<link>'s imagesrcset and imagesizes attributes:
```
martinthomson, line 1: I'm constantly amazed at YAML and its ability to interpret what would be garbage input in other contexts.  I assume that this is machine-generated.

gh-data-summary.json
```diff
@@ -0,0 +1,9731 @@
+[
```
martinthomson, line 1: Does this file need to be checked in to the repository?

martinthomson (commented):
Looking forward to this work being ready.

Are you planning to add a workflow for this?

activities.yml
zcorpan: It is yes (`json2yaml.py`)

gh-data.py
zcorpan: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users says 60 requests per hour, and currently this will make 12 requests. (Could make it avoid the last request by looking at the `link` response header.)

If we make a GH Action workflow to batch updates if the most recent run was less than 15 minutes ago or so (which seems like a good idea anyway), unauthenticated shouldn't hit the rate limit for now.

zcorpan: I've experimented in https://github.com/zcorpan/gh-actions-test to get such a setup working (with help of ChatGPT 4o). The workflow will push built files to `gh-pages`, which I think is useful to be able to see the history.

gh-data-summary.json
zcorpan: No, at least not to `main`. My plan is to check in generated files to `gh-pages` with a GH Action.

tantek (approved):
Agreed with this direction and simplification.

zcorpan:
This should be ready now, except the GH Actions need a deploy key and environment secret for this repo which I can't add currently.

zcorpan:
> except the GH Actions need a deploy key and environment secret for this repo which I can't add currently.

This is now done.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b gh-pages
+          git add -f gh-data.json
+          git add -f gh-data-summary.json
+          git add -f merged-data.json
+          git add .
+          git commit -m "Update GitHub Pages with latest data"
+          git remote set-url origin git@github.com:${{ github.repository }}.git
+          git push origin gh-pages --force
```
martinthomson, line 54: A force push seems unnecessary here.  Better to let it fail, I'd say.

Alternatively, you could use the GitHub pages action, like I do [here](https://github.com/martinthomson/asciitable/blob/main/.github/workflows/index.yml).  The advantage of that is that you don't end up tracking changes constantly.

DEVELOP.md
```diff
@@ -0,0 +1,43 @@
...
+
+Then load http://localhost:8000/ in a web browser.
+
+## GitHub Actions / Publishing
+
+Publishing happens automatically when pushing to `main` or when changing labels in GitHub.
+
+The latter runs the `labels-change.yml` workflow directly and then the `schedule-labels-change.yml` workflow periodically checks the history of workflow jobs to determine if labels have changed,
+and if so pushes an empty commit to the `trigger-build` branch,
+which causes the `build-and-deploy.yml` workflow to run.
```
martinthomson, line 40: This seems a little convoluted.  I can see why you want to avoid running a build on every label change, but this does exactly that.  And I can't see why you have a schedule-labels-change and a build-and-deploy as separate things.

I would have thought that you could just use the API to query for recent changes to issues, filtered for label changes.

martinthomson (approved):
This certainly seems like an improvement.

I have a few questions about the complex deployment architecture here, but I assume that most of this is there for good reasons (that I don't understand).  I'm OK with shipping this.

DEVELOP.md
zcorpan: Hmm, indeed a job is triggered on every label change. I can look into your suggested alternative, I didn't think of that.

Checking for label changes and building is separate because building should happen on changes to `main` even if there were no label changes. Maybe that can be done from a single workflow, though (with `github.event_name`).

larseggert:
Not sure if this should be part of this PR, but it might be a good idea to enable `dependabot` and `actionlint` for this repo?

bgrins:
I'm good with the change, pending any updates to the actions that you're looking into.

bgrins (approved):

DEVELOP.md
zcorpan: As far as I can tell you can only get label changes for a single issue at a time, so this would be somewhat complex also. I have now simplified by running `build-and-deploy.yml` on a cron job (every 6 hours), and rely on `git commit` and `git push` not doing anything if there are no changes.

.github/workflows/build-and-deploy.yml
zcorpan: Now only `merged-data.json` is added.

Issue #522, opened 2021-05-07
Request for position: JPEG XL


saschanaz:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: JPEG XL Image Coding System
* Specification or proposal URL:
  * ISO standards: 
    * Part 1: https://www.iso.org/standard/77977.html
    * Part 2: https://www.iso.org/standard/80617.html
    * Part 3: https://www.iso.org/standard/80618.html
    * Part 4: https://www.iso.org/standard/80619.html
  * Project page: https://gitlab.com/wg1/jpeg-xl
* Caniuse.com URL (optional): https://caniuse.com/jpegxl
* Bugzilla URL (optional): https://bugzil.la/jpeg-xl
* Mozillians who can provide input (optional): 

### Other information

[This is included in Nightly 90](https://bugzilla.mozilla.org/show_bug.cgi?id=1707590) but there hasn't been an active discussion about whether this is the right way to go.

hsivonen:
I think JXL is, in a good way, unlike multiple other new potentially-lossy image formats in three key ways:

1. Can recompress JPEG without further visual changes. (I think this is a big deal for migration path, since it allows size wins without human supervision for quality loss. Also, this gives the format a more legitimate claim to a "JPEG something" brand than e.g. JPEG2000. It should be pretty non-controversial that this property can be verified, though I haven't personally done the verification.)
2. Is designed for high-fidelity still photos. (The primary way both JPEG2000 and the video codec-based image formats improve on JPEG is that they have non-awful but still visible artifacts at compression ratios where JPEG has very awful and very visible artifacts. I think it's important to also address the case where you don't want visible artifacts but still want a substantial compression win over JPEG. How well JXL meets its design intent on this point is harder to assess than the previous point. I haven't done this assessment personally, either, but I react favorably to this even being the design intent.)
3. Less generation loss, so better suited for authoring workflow. (I haven't verified this personally)
4. Better progressive decoding story: Does not need a site-provided mechanism for low-res placeholder swapping if the browser implements incremental painting of incremental decode.
5. No tile boundaries based on maximum video frame size. (Unclear if at all relevant at high-fidelity compression ratios.)

Unless a Mozilla expert shows that key marketing claims are wrong, I think point 1 alone is such a big deal that we should take a favorable position on the format on point 1 alone, and, personally, I've been hoping that Firefox would add support.

[...]

PR #1064, opened 2024-09-03
Firefox will consider a Rust implementation of JPEG-XL

bholley:
Over the past few months, we’ve had some productive conversations with the JPEG-XL team at Google Research around the future of the format in Firefox. Our primary concern has long been the increased attack surface of the reference decoder (currently behind a pref in Firefox Nightly), which weighs in at more than 100,000 lines of multithreaded C++. To address this concern, the team at Google has agreed to apply their subject matter expertise to build a safe, performant, compact, and compatible JPEG-XL decoder in Rust, and integrate this decoder into Firefox. If they successfully contribute an implementation that satisfies these properties and meets our normal production requirements, we would ship it.

Time will tell whether the format succeeds in becoming a universal JPEG replacement in the way some folks hope. In the event that it does, it would be unfortunate to potentially introduce memory safety vulnerabilities across the myriad of applications that would eventually need to support it. A safe, fast, and battle-tested Rust decoder from the original team could make that scenario much less likely, and so we’re using our leverage to encourage progress on this front.

Previous discussion in #522.

### DIFF
--- a/activities.json
+++ b/activities.json
@@ -955,7 +955,7 @@
     "mdnUrl": null,
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1539075",
     "mozPosition": "neutral",
-    "mozPositionDetail": "JPEG-XL includes features and performance that might differentiate it from other formats, but the benefits it provides are not significant enough on their own to justify the cost of adding another C++ image decoder to Firefox. A memory-safe decoder would reduce these costs considerably, and we are open to shipping one that meets our requirements.",
+    "mozPositionDetail": "JPEG-XL includes features and performance that might differentiate it from other formats, but the benefits it provides are not significant enough on their own to justify the cost of adding another C++ image decoder to browsers. A memory-safe decoder would reduce these costs considerably, and we are open to shipping one that meets our requirements.",
     "mozPositionIssue": 522,
     "org": "Other",
     "title": "JPEG-XL",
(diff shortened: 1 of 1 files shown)


activities.json
```diff
@@ -955,7 +955,7 @@
     "mdnUrl": null,
     "mozBugUrl": "https://bugzilla.mozilla.org/show_bug.cgi?id=1539075",
     "mozPosition": "neutral",
-    "mozPositionDetail": "JPEG-XL includes features and performance that might differentiate it from other formats, but the benefits it provides are not significant enough on their own to justify the cost of adding another raster image format to the Web.",
+    "mozPositionDetail": "JPEG-XL includes features and performance that might differentiate it from other formats, but the benefits it provides are not significant enough on their own to justify the cost of adding another C++ image decoder to Firefox. A memory-safe decoder would reduce these costs considerably, and we are open to shipping one that meets our requirements.",
```
martinthomson, line 958:
```suggestion
    "mozPositionDetail": "JPEG-XL includes features and performance that might differentiate it from other formats, but the benefits it provides are not significant enough on their own to justify the cost of adding another C++ image decoder to browsers. A memory-safe decoder would reduce these costs considerably, and we are open to shipping one that meets our requirements.",
```

I think that we can make a broader statement safely, even if it is just Firefox/Gecko that is making the statement :)

martinthomson (approved):
A take-it-or-leave-it comment, but this looks good to me.

tantek (approved):
LGTM and I concur with @martinthomson's edit.

PR #1099, opened 2024-10-15
Deduplicate GitHub issues in activities.json

### LINKED ISSUE #1087: Use separate issues for separate standards positions
The following issues are referenced by more than one item in `activities.json`, which is incompatible with https://github.com/mozilla/standards-positions/pull/1063 . I will split them up by filing new issues.

- #93
  - CSS Layout API
  - CSS Painting API
  - CSS Properties and Values API
  - CSS Typed OM
  - Worklets
- #60 
  - Custom elements
  - HTML Imports
  - Shadow trees (formerly known as Shadow DOM)
  - template element
- #27 
  - Web Share API
  - Web Share Target API

zcorpan:
Fixes #1087


tantek (changes requested):
The mozPositionIssue updates all look correct (and finer-grained) to me. Thanks for doing this.

The only nit I noticed (unchanged from before, but worth fixing while touching these since separate issues with descriptions got created) is a couple of instances of:

> "mozPositionDetail": "A welcome successor to XBL!",

Could these be replaced with something slightly more specific to the technology being reviewed? (just as the issue numbers are more specific now)

smaug----:
Maybe "Standardized successor of XBL" would be enough?

(Custom elements and Shadow tree aren't really better than XBL, but there is at least good specifications for both.)

tantek (approved):
Would have preferred text particular to each spec (like what does it provide that replaces what XBL functionality) but better to land this clean up and improve that later.

Issue #618, opened 2022-03-10
FedCM (was WebID)


samuelgoto:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: [Federated Credential Management API](https://github.com/fedidcg/FedCM)
* Specification or proposal URL: 
    * [explainer](https://github.com/fedidcg/FedCM/blob/main/explainer.md)
    * [specification](https://fedidcg.github.io/FedCM/)
    * [TPAC2021](https://watch.videodelivery.net/29bab61e04e8cabf1517e5885c9fe4cf) presentation (good introduction)
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): 
    * Since we started, we've discussed this proposal with a few mozillians: 
    * stpeter (general point of contact)
    * @martinthomson (overall guidance)
    * tvyas, johannhof (early privacy discussions)
    * danmills, rfkelly, arthur (on mozilla personas)

### Other information

Various other reviews:

* [WICG thread](https://discourse.wicg.io/t/proposal-webid-privacy-preserving-federated-sign-in-api/4661)
* [Federation and Browsers Workshop](https://github.com/fedidcg/FedCM/blob/main/meetings/2021/25-26_May_2021.md)
* [Intent To Prototype](https://groups.google.com/a/chromium.org/g/blink-dev/c/2B4TJ7j2U4M/m/1X5T3OszCAAJ)
* [Intent To Experiment](https://groups.google.com/a/chromium.org/g/blink-dev/c/kws-gltC5us)
* [FedID CG](https://www.w3.org/community/fed-id/)
* early [TAG review](https://github.com/w3ctag/design-reviews/issues/622)
* [Ready For Trial](https://groups.google.com/a/chromium.org/g/blink-dev/c/jlV_1m7uUAg)
* spec [TAG review](https://github.com/w3ctag/design-reviews/issues/718)
* [WebKIt's Position Thread](https://lists.webkit.org/pipermail/webkit-dev/2022-March/032153.html)

martinthomson:
So we've been quiet on this one, but @samuelgoto knows that we've been working out what to do with it for some time.

We're broadly supportive of doing this work, even though we've identified a number of issues that mean that this is not an unreserved recommendation.

Overall, being able to restore some amount of user control over the cross-site flow of information about them is one of our most important efforts.  This covers a number of different facets, but we think that something like FedCM could be an important part of the overall strategy.

[...]

PR #1104, opened 2024-10-31
FedCM Update

bvandersloot-mozilla:
It's been a couple years. This should be a good update.

@martinthomson is a good person to review and merge!

### DIFF
--- a/activities.yml
+++ b/activities.yml
@@ -817,28 +817,29 @@ Federated Credential Management API:
   id: fedcm
   issue: 618
   mdn: null
-  position: positive
-  rationale: 'Federated login is a widely-used feature on the web with significant
-    user benefits in usability and security.  Unfortunately, federated identity on
-    the web relies on the same techniques that are used to track web users.  The Federated
-    Credential Management API puts the browser in control of managing cross-site logins.  Browsers
-    can use this API as a way to give web users better ability to control and monitor
-    how their identity - and any information related to their identity - is exchanged
-    between sites.  Including the browser in a mediating role will adversely affect
-    some cross-site interactions, in some cases making them less efficient or even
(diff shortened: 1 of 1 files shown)


tantek (changes requested):
I checked the linked issue https://github.com/mozilla/standards-positions/issues/618 and didn't see any discussion regarding the change in position from "positive" to "neutral". Could you add a comment there with the updated analysis (or link if elsewhere) that led to the new "neutral" position?

bvandersloot-mozilla:
@tantek: I've added a comment with some updated analysis in #618. I also updated the URL so it doesn't rely on a redirect. It recently published a FPWD in a WG, but has a lot of issues to close before it becomes a CR. Should I leave the org as "Proposal" or update to "W3C"?

activities.json
```diff
@@ -654,12 +654,12 @@
...
     "mozPositionIssue": 618,
     "org": "Proposal",
```
tantek, line 660:
```suggestion
    "org": "W3C",
```

tantek (changes requested):
yes, change org to W3C, and similarly, if you could update the mozPositionDetail with a summary of the comment you made on #618 that would help provide a self-contained update of why we are changing the position to "neutral". Otherwise from an outside perspective it's not obvious (without hunting/clicking) why position changed.

zcorpan:
I've rebased this so that it edits activities.yml instead, and added "W3C" per @tantek 's comment. I haven't changed the rationale as @tantek requested, leaving that to @bvandersloot-mozilla

bvandersloot-mozilla:
I have already updated `rationale` with a summary of that comment. `mozPositionDetail` got refactored to that IIRC.

tantek (approved):
LGTM. Thanks for this critical update.

PR #1047, opened 2024-06-27
Positive on the Cookie Store API (subset)

jesup:

Issue #241, opened 2020-01-09
`timezonechange` event


mathiasbynens:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: `timezonechange` event
* Specification or proposal URL: https://github.com/whatwg/html/pull/3047
* Caniuse.com URL (optional): 
* Bugzilla URL (optional): 
* Mozillians who can provide input (optional): @annevk, @chrisdavidmills

### Other information

Mozilla's helpful involvement on the spec patch seems to imply some level of support, but it would be great to get that explicitly confirmed or denied. Is Mozilla supportive of the proposed `timezonechange` event?

dbaron:
Also cc @jswalden for thoughts on implementation feasibility.

annevk:
To be clear, there's two things to evaluate:

1. The ability to dispatch an event when the timezone changes.
2. To not expose a timezone change to script until the event is dispatched. I.e., store the timezone as internal state and update it during the same task that dispatches the event.

1 satisfies the use case, 2 would be ideal and also matches existing APIs that dispatch events for (external) state changes.

dbaron:
Jeff Yaskin [pointed out](https://groups.google.com/a/chromium.org/d/msg/blink-dev/4BUSE2aTQEc/a5-gBNTdCAAJ) an interesting concern:

> Nick Doty pointed out that if this event fires at the same time across
> all open tabs, the timestamp of that event becomes a cross-origin
> tracking vector. I believe this is fixed if the event is delayed until
> the user next activates that tab. If that much delay isn't acceptable,
> there may be a way to fire the events earlier if they're spread across
> enough time, but I don't immediately know how much spread is needed. 

but it sounds like this is being looked into.

dbaron:
So I think given that the discussion of how to handle correlation of the event across tabs hasn't been resolved yet, I think we should take a `defer` position on this one and come back to it once that is addressed.

mathiasbynens:
> So I think given that the discussion of how to handle correlation of the event across tabs hasn't been resolved yet

It seems there is now general agreement on the upstream thread on how to resolve these issues, e.g. https://github.com/whatwg/html/pull/3047#issuecomment-573847868 & https://github.com/whatwg/html/pull/3047#issuecomment-663181802. Does Mozilla agree with this?

Can we reopen this issue?

annevk:
[...]

PR #1106, opened 2024-11-05
Add a link to issue WICG/cookie-store #241 about the Cookie-Store API subset

bakulf:
See https://github.com/mozilla/standards-positions/pull/1047#pullrequestreview-2146610462

This PR adds a link to the issue WICG/cookie-store #241 about the Cookie-Store API subset


zcorpan:
I've added the link in #1047

PR #1123, opened 2024-11-20
Fix deploy workflow to fetch and rebase

zcorpan:
The workflow fails to push changes to gh-pages because it has commits that are not in the local branch.


tantek (approved):
LGTM

PR #1130, opened 2024-11-27
Extract the first URL in gh-data.py

zcorpan:
Issue descriptions sometimes have text around URLs.

### DIFF
--- a/gh-data.py
+++ b/gh-data.py
@@ -66,6 +66,15 @@ def process_labels(labels):
     }
 
 
+def get_url(text):
+    m = re.search(r"\b(https?://\S+)", text)
+    if m:
+        url = m.group()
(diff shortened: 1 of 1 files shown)


gh-data.py
```diff
@@ -66,6 +66,15 @@ def process_labels(labels):
     }
 
 
+def get_url(text):
+    m = re.search(r"\b(https?://\S+)", text)
+    if m:
+        url = m.group()
+        if url.endswith(','):
+            url = url[:-1]
```
martinthomson, line 74: Do you need to consider `[](https://example.com)` as well?

zcorpan: Yes, good point.

tantek (approved):

PR #1132, opened 2024-12-02
Fix deploy workflow (take 4)

zcorpan:
If there is nothing to commit for merged-data.json, git commit exits with a non-zero code.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -65,7 +65,11 @@ jobs:
           git add -f merged-data.json
 
           # Commit changes if there are any
-          git commit -m "Update GitHub Pages with latest data"
+          if ! git diff --cached --quiet; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -65,7 +65,11 @@ jobs:
           git add -f merged-data.json
 
           # Commit changes if there are any
-          git commit -m "Update GitHub Pages with latest data"
+          if ! git diff --cached --quiet; then
```
tantek, line 68:
```suggestion
          if git diff --cached --quiet; then
```

tantek (changes requested):
Looks like the logic is reversed. --quiet implies --exit-code which returns 1 for changes, 0 for no changes per https://git-scm.com/docs/git-diff

tantek (approved):

PR #1134, opened 2024-12-03
Use aria-describedby on <summary>

zcorpan:
It can be unclear for a non-visual user which row the details element is for.


zcorpan:
This is to address feedback from @jcsteh.

tantek (approved):

PR #1140, opened 2024-12-06
Use YAML literal block syntax for description, rationale

zcorpan:
This prevents some syntax issues with YAML, e.g. '#' can be used in a literal block without being treated as a line comment.

The switch from PyYAML to ruamel.yaml was needed to validate the syntax style.

### DIFF
--- a/activities.py
+++ b/activities.py
@@ -1,150 +1,136 @@
-import yaml
-import re
-import sys
+from ruamel.yaml import YAML
+from ruamel.yaml.comments import CommentedMap
+from ruamel.yaml.scalarstring import LiteralScalarString
 import argparse
 import requests
-from typing import Any, Dict, List, Tuple
+import sys
+from typing import Any, Dict
+
 
 class YAMLValidator:
     def __init__(self, file_path: str, fix: bool = False):
         self.file_path = file_path
         self.fix = fix
-        self.issues = set()
         self.errors = []
         self.modified_data = None
+        self.yaml = YAML()
+        self.yaml.preserve_quotes = True
 
-    def load_yaml(self) -> Tuple[Dict[str, Any], Dict[int, str]]:
+    def load_yaml(self) -> CommentedMap:
         try:
             with open(self.file_path, 'r') as file:
-                lines = file.readlines()
-                data = yaml.safe_load("".join(lines))
-                line_map = {i + 1: line.strip() for i, line in enumerate(lines)}
-                return data, line_map
-        except yaml.YAMLError as e:
-            raise ValueError("Failed to load YAML: Invalid YAML format") from e
+                data = self.yaml.load(file)
+                return data
+        except Exception as e:
+            raise ValueError(f"Failed to load YAML: {e}") from e
 
-    def log_error(self, message: str, key: str = "", line: int = None, fixable=False):
-        tofix = " (Use --fix to fix.)" if fixable else ""
-        if line is not None:
-            self.errors.append(f"Error at line {line} in key '{key}': {message}{tofix}")
-        else:
-            self.errors.append(f"Error: {message}{tofix}")
+    def log_error(self, message: str, key: str = ""):
+        self.errors.append(f"Error in key '{key}': {message}")
 
-    def validate_top_level_keys(self, data: Dict[str, Any], line_map: Dict[int, str]):
+    def validate_top_level_keys(self, data: CommentedMap):
         keys = list(data.keys())
-        modified = False
-
-        # Check for non-string keys
-        non_string_keys = [key for key in keys if not isinstance(key, str)]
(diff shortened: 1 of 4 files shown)


activities.py
```diff
@@ -1,150 +1,136 @@
...
-            self.log_error(f"'id' must be a string without whitespace, got {item['id']}", key, line)
-        if 'description' in item and item['description'] not in [None, '']:
-            if not isinstance(item['description'], str) or not item['description'].strip():
-                self.log_error(f"'description' must be a string or null, got {item['description']}", key, line)
+            self.log_error(f"'id' must be a string without whitespace, got {item['id']}.", key)
+
         if 'bug' in item and item['bug'] not in [None, '']:
             if not isinstance(item['bug'], str) or not item['bug'].startswith("https://bugzilla.mozilla.org/show_bug.cgi?id="):
-                self.log_error(f"'bug' must be a URL starting with 'https://bugzilla.mozilla.org/show_bug.cgi?id=', got {item['bug']}", key, line)
+                self.log_error(f"'bug' must be a URL starting with 'https://bugzilla.mozilla.org/show_bug.cgi?id=', got {item['bug']}.", key)
```
martinthomson, line 85: If we're requiring this, why not include just the bug number (and take the `https://bugzilla...` as implied?)

Not for this change, but as a followup perhaps.

activities.py
```diff
@@ -1,150 +1,136 @@
...
+
         if 'caniuse' in item and item['caniuse'] not in [None, '']:
             if not isinstance(item['caniuse'], str) or not item['caniuse'].startswith("https://caniuse.com/"):
-                self.log_error(f"'caniuse' must be a URL starting with 'https://caniuse.com/' or empty, got {item['caniuse']}", key, line)
+                self.log_error(f"'caniuse' must be a URL starting with 'https://caniuse.com/', got {item['caniuse']}.", key)
```
martinthomson, line 89: As above.

activities.py
```diff
@@ -1,150 +1,136 @@
...
+
+        if 'mdn' in item and item['mdn'] not in [None, '']:
+            if not isinstance(item['mdn'], str) or not item['mdn'].startswith("https://developer.mozilla.org/en-US/"):
+                self.log_error(f"'mdn' must be a URL starting with 'https://developer.mozilla.org/en-US/', got {item['mdn']}.", key)
```
martinthomson, line 93: As above

activities.py
```diff
@@ -1,150 +1,136 @@
...
+
+    def validate_literal_block(self, data: CommentedMap, key: str, field: str):
+        value = data.get(key, {}).get(field)
+        if value is not None:
+            if not isinstance(value, LiteralScalarString):
+                if self.fix:
+                    data[key][field] = LiteralScalarString(value)
+                else:
+                    self.log_error(f"'{field}' must use literal block syntax (|).", key)
+            elif not value.endswith('\n'):
```
martinthomson, line 46: Is it even possible to have a value that doesn't have a trailing `\n`?

activities.yml
```diff
@@ -1,2989 +1,2479 @@
 <link>'s imagesrcset and imagesizes attributes:
   bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1515760
-  caniuse: null
-  description: Adds imagesrcset and imagesizes attributes to &lt;link> which correspond
-    to the srcset and sizes attributes of &lt;img> respectively, for the purposes
-    of preloading.
+  caniuse:
```
martinthomson, line 3: We've a bunch of these empty values.  I assume that these can be dropped from the file safely.  Should they be?

merge-data.py
```diff
@@ -13,11 +17,19 @@
...
-        if dict1[key]:
-            dict2[key] = dict1[key]
-    dict2.pop('issue', None)
+    """
+    Merges data from dict1 into dict2. Keys in dict1 take precedence if not None.
+    Strips trailing newline from 'description' and 'rationale'.
+    """
+    for key, value in dict1.items():
+        if value is not None:
+            if key in ['description', 'rationale'] and isinstance(value, str):
```
martinthomson, line 26: Do you need to check for description/rationale?  Could this trim all strings?

martinthomson (approved):

activities.py
zcorpan: Yes:
```
foo: |-
  bar
```

activities.py
zcorpan: I think it's easier for contributors to paste the full URL for all of the URLs than to find out what substring should be used. Though new issues will have the URLs in the issue OP instead, so it doesn't matter so much.

merge-data.py
zcorpan: It could trim all strings.

PR #1141, opened 2024-12-09
Update position for NEL (positive)

### LINKED ISSUE #99: Network Error Logging (NEL)

* Specification Title: Network Error Logging (NEL)
* Specification or proposal URL: https://w3c.github.io/network-error-logging/
* Chromestatus.com URL (optional): https://www.chromestatus.com/feature/5391249376804864
* Mozillians who can provide input (optional): Patrick McManaus

zcorpan:
Closes #99.

### DIFF
--- a/activities.yml
+++ b/activities.yml
@@ -1498,11 +1498,9 @@ Network Error Logging:
   id: network-error-logging
   issue: 99
   mdn: https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging
-  position: negative
-  rationale: The API enables the collection of user-specific information that sites
-    might not otherwise be able to observe, which includes information that might
-    be private under some circumstances. Furthermore, the specification does not seem
-    to track changes to the Reporting API it builds upon and seems effectively unmaintained.
+  position: positive
+  rationale: |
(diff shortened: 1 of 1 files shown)


valenting (approved):

tantek (changes requested):
Two problems I can see with this update having read the comment thread on #99:


1: should (likely) be overall 'neutral' not 'positive'.

I don't think "subset of spec is positive" is sufficient for a "positive" evaluation overall, but rather "neutral" at best if the positive at least somewhat offsets the negative.

Have we defined what that "subset" is?

Is there an issue open to reduce the spec to that subset?

If the spec is reduced to the positive subset, then we can re-evaluate as "positive".

This is something we should probably capture in a guide for s-p.


2: original evaluators/critics did not agree to change

Also unusual is that the flipping of position was not commented on or re-evaluated by any of the people who proposed the prior (negative) position.

E.g. I would want to at least see @mozfreddyb explicitly say he also thinks that that same subset is positive.

Even with that though, I would still conclude "neutral" per (1), unless Freddy was convinced that the subset is so positive that it overwhelming overrides remaining negative aspects of the spec (which btw, no one, not even @valenting's comment https://github.com/mozilla/standards-positions/issues/99#issuecomment-2459066504, has addressed).


"Some parts of this spec are positive" has NEVER been justification for a "positive" position. 

Even changing this to "neutral" would require explicit reasoning and explanations why & how the benefits from the positive aspects sufficiently override the harms from the negative aspects.

zcorpan:
Closing this for now per https://github.com/mozilla/standards-positions/issues/99#issuecomment-2663506851

PR #1145, opened 2024-12-16
Visually hide the text 'Details'

zcorpan:
This text made it harder to scan the specification links and took up space.


zcorpan:
Before:
Screenshot before, with the text 'Details'

After:
Screenshot after, with only disclosure triangles

martinthomson (approved):

PR #1147, opened 2024-12-18
Use markdown instead of HTML in activities.yml

zcorpan:
Also support code in the spec link text.

### DIFF
--- a/index.html
+++ b/index.html
@@ -212,8 +212,6 @@
        const fragment = document.createDocumentFragment();
        const codePattern = /`([^`]+)`/;
 
-       let cursor = 0;
-
        while (true) {
          const match = codePattern.exec(markdown);
          if (!match) {
@@ -224,7 +222,7 @@
            break;
          }
 
-         const textBefore = markdown.slice(cursor, match.index);
+         const textBefore = markdown.slice(0, match.index);
          if (textBefore) {
            fragment.appendChild(document.createTextNode(textBefore));
          }
@@ -234,12 +232,8 @@
          codeElement.textContent = match[1];
          fragment.appendChild(codeElement);
 
-         // Move our cursor beyond this match
-         cursor = match.index + match[0].length;
          // Slice off what we've already processed
-         markdown = markdown.slice(cursor);
-         // Reset cursor for next iteration
-         cursor = 0;
+         markdown = markdown.slice(match.index + match[0].length);
        }
 
        return fragment;
@@ -248,8 +242,6 @@
      function expandLinks(markdown) {
        const fragment = document.createDocumentFragment();
        const linkPattern = /\[([^\]]+)\]\(((?:#|https:\/\/)[^\)]+)\)/;
-
-       let cursor = 0;
 
        while (true) {
          const match = linkPattern.exec(markdown);
@@ -262,7 +254,7 @@
          }
 
          // Text before the link
-         const textBefore = markdown.slice(cursor, match.index);
+         const textBefore = markdown.slice(0, match.index);
          if (textBefore) {
            fragment.appendChild(expandInline(textBefore));
          }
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -207,36 +208,47 @@ <h2>Legend</h2>
...
-       const template = document.createElement('template');
-       const div = template.content.ownerDocument.createElement('div');
-       template.content.append(div); // force parser context to "in body"
-       div.innerHTML = value;
-       const iterator = document.createNodeIterator(div);
-       let node = iterator.nextNode(); // This is the div
-       while (node = iterator.nextNode()) {
-         if (node.nodeType === Node.TEXT_NODE) {
-           continue;
+     function parseMarkdown(markdown, enableLinks = true) {
```
martinthomson, line 211: I'm not a fan of defaults being true.  Though see below for an alternative structure.

index.html
```diff
@@ -207,36 +208,47 @@ <h2>Legend</h2>
...
+       // Create a document fragment to hold the parsed elements
+       const fragment = document.createDocumentFragment();
+
+       // Regex for either code or link (only # or https://)
+       const markdownPattern = enableLinks ? /(`([^`]+)`|\[([^\]]+)\]\(((?:#|https:\/\/)[^\)]+)\))/g : /(`([^`]+)`)/g;
```
martinthomson, line 216: It looks like you have two obvious phases here: link parsing and code parsing.  Could that be two separate passes?  That might remove the need for the argument as well.

```js
function expandLinks(md) {
  // match then expandInline in the text chunks before appending those.
}
function expandInline(md) {
  // match /`([^`]+)`/ and maybe /\b*([^*]+)*\b/ as well.
}
```

martinthomson: Separately, you don't need the '/g' if you are parsing in chunks like this.

index.html
```diff
@@ -207,36 +208,81 @@ <h2>Legend</h2>
...
+         // Reset cursor for next iteration
+         cursor = 0;
+       }
+
+       return fragment;
+     }
+
+     function expandLinks(markdown) {
+       const fragment = document.createDocumentFragment();
+       const linkPattern = /\[([^\]]+)\]\(((?:#|https:\/\/)[^\)]+)\)/;
```
martinthomson, line 250: Do you want to match on bare links as well?  Like `/\b(https:\/\/\S+)\b/`?  (Save that for later perhaps.)

index.html
```diff
@@ -207,36 +208,81 @@ <h2>Legend</h2>
...
-       const iterator = document.createNodeIterator(div);
-       let node = iterator.nextNode(); // This is the div
-       while (node = iterator.nextNode()) {
-         if (node.nodeType === Node.TEXT_NODE) {
-           continue;
+     function expandInline(markdown) {
+       const fragment = document.createDocumentFragment();
+       const codePattern = /`([^`]+)`/;
+
+       let cursor = 0;
```
martinthomson, line 215: I don't think that you are using this variable in the code here, see the one suggestion below.  (Same for the link expansion code)

index.html
```diff
@@ -207,36 +208,81 @@ <h2>Legend</h2>
...
+         const codeElement = document.createElement('code');
+         codeElement.textContent = match[1];
+         fragment.appendChild(codeElement);
+
+         // Move our cursor beyond this match
+         cursor = match.index + match[0].length;
+         // Slice off what we've already processed
+         markdown = markdown.slice(cursor);
+         // Reset cursor for next iteration
+         cursor = 0;
```
martinthomson, line 242:
```suggestion
         // Slice off what we just matched
         markdown = markdown.slice(match.index + match[0].length);
```

martinthomson (approved):
Just one actual comment, which will save a few lines.

index.html
zcorpan: No, the data should have links marked up.

PR #1152, opened 2024-12-20
Add table sorting

### LINKED ISSUE #1120: Sort entries on the dashboard
The new dashboard is great, but the items in the table seem to be arbitrarily ordered.

zcorpan:
Fixes #1120.

### DIFF
--- a/index.html
+++ b/index.html
@@ -98,9 +98,9 @@
      td.more-info :any-link:not([title="Discussion"]) { padding: 0; text-decoration: none; }
      td.link :any-link { padding: 0.25rem; }
      th.concerns { width: 7em; }
-     th.position { width: 4em; }
+     th.position { width: 4.25em; }
      th.topics { width: 3.5em; }
-     th.venues { width: 3.5em; }
+     th.venues { width: 4em; }
      th.more-info { width: 9.5em; }
      th, td { padding: 0.5rem; }
      th.link { width: 3em; text-align: center; }
@@ -126,6 +126,14 @@
      span.negative { background-color: var(--negative-bg-color); color: var(--negative-color); }
      span.neutral { border-color: currentColor; }
      span.defer, span.under-consideration { background-color: var(--defer-bg-color); color: var(--defer-color); }
+
+     /* sorting */
+     th.sortable { padding-top: 0; padding-bottom: 0; }
+     th button { appearance: none; padding: 0.25em 0; font: inherit; display: inline; background-color: transparent; border: none; cursor: default; user-select: none; }
+     th button::after { font-size: 0.8125em; content: " ▼"; opacity: 0.3; }
+     th button:hover::after, th button:focus::after { opacity: 0.6 }
+     th[aria-sort] button::after { opacity: 1; }
+     th[aria-sort="descending"] button::after { content: " ▲"; }
     </style>
 
     <script>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -439,6 +447,69 @@ <h2>Legend</h2>
...
+         button.textContent = header.textContent;
+         header.setAttribute("aria-label", header.textContent);
+         header.textContent = '';
+         header.append(button);
+         if (isInitDescending) {
+           header.setAttribute("aria-sort", "descending");
+         }
+
+         button.onclick = () => {
+           const isAscending = header.getAttribute("aria-sort") !== "ascending";
```
martinthomson, line 473: This is a pretty confusing assignment.  I think that your intent is to say that the sort should be ascending when the current state is not ascending, but the negation and naming here makes it look backwards.  Maybe:

```suggestion
           const sortAscending = header.getAttribute("aria-sort") !== "ascending";
```

index.html
```diff
@@ -439,6 +447,69 @@ <h2>Legend</h2>
...
+         if (a === "" || b === "") {
+           if (a === "" && b === "") {
+             return 0;
+           }
+           return a === "" ? 1 : -1;
+         }
+
+         return ascending
+           ? a.localeCompare(b, undefined, { numeric: sortNumerically })
+           : b.localeCompare(a, undefined, { numeric: sortNumerically });
```
martinthomson, line 501:
```suggestion
         [a, b] = ascending ? [a, b] : [b, a];
         return a.localeCompare(b, undefined, { numeric: sortNumerically });
```

index.html
```diff
@@ -439,6 +447,69 @@ <h2>Legend</h2>
...
+         const isInitDescending = header.classList.contains("more-info");
+         const sortNumerically = isInitDescending;
+         button.setAttribute("aria-label", `Sort by ${header.textContent}`);
+         button.textContent = header.textContent;
+         header.setAttribute("aria-label", header.textContent);
+         header.textContent = '';
+         header.append(button);
+         if (isInitDescending) {
+           header.setAttribute("aria-sort", "descending");
+         }
```
martinthomson, line 470: From the code, it wasn't clear to me that this column was:
a) reverse-sorted to begin with
b) effectively the primary sort column for the table

A comment to that effect might help.

index.html
```diff
@@ -439,6 +447,69 @@ <h2>Legend</h2>
...
+
+       const headers = table.querySelectorAll("th");
+
+       headers.forEach((header, columnIndex) => {
+         if (!header.classList.contains('sortable')) {
+           return;
+         }
+         const button = document.createElement("button");
+         const isInitDescending = header.classList.contains("more-info");
+         const sortNumerically = isInitDescending;
```
martinthomson, line 462:
```suggestion
         const sortNumerically = isInitDescending;  // Coincidentally, this is the only numeric column.
```

index.html
```diff
@@ -439,6 +447,69 @@ <h2>Legend</h2>
...
+       const table = document.querySelector("table");
+
+       const headers = table.querySelectorAll("th");
+
+       headers.forEach((header, columnIndex) => {
+         if (!header.classList.contains('sortable')) {
+           return;
+         }
+         const button = document.createElement("button");
+         const isInitDescending = header.classList.contains("more-info");
```
martinthomson, line 461:
```suggestion
         // "more-info" is the column that starts out being sorted; in reverse order.
         const isInitDescending = header.classList.contains("more-info");
```

martinthomson (approved):

index.html
zcorpan: Thanks, fixed.

PR #1180, opened 2025-02-17
Fix links to 'null'

zcorpan:
Link to the explainer URL instead if spec URL is null, or omit href if both are null.


tantek (approved):
LGTM

Issue #94, opened 2018-06-25
Cookie-Store API


bakulf:
## Request for Mozilla Position on an Emerging Web Specification

* Specification Title: Cookie Store API
* Specification or proposal URL: https://wicg.github.io/cookie-store/

marcoscaceres:
@bakulf, anyone in particular you'd like to tag to review the spec and provide feedback?

bakulf:
The spec is still in an early state and many sections are unwritten yet.
But at the same time we have a good WPT coverage.
I wrote some code to implement part of this spec, but I wonder if somebody else has feedback.

annevk:
See https://github.com/w3ctag/design-reviews/issues/290#issuecomment-394959991 for some high-level concerns.

dbaron:
/cc @ddragana @martinthomson @bzbarsky for additional feedback on this.

The goals stated in [the explainer](https://wicg.github.io/cookie-store/explainer.html) seem reasonable to me -- though I suppose a counterargument to them is that we'd like cookies to disappear and having them *not* be available on workers helps with that.  (Unfortunately, in reality, having them have poor performance characteristics probably doesn't help with that.)

@annevk, I'm not sure I understand what your concerns are.  Do they imply opposition to the existence of the API or concerns about the details of how it's being done?

martinthomson:
Over time I would like to see old janky APIs go away, so I don't think that this is terrible.  I have a hard time getting excited about implementing it - it's the sort of long-term work that doesn't have any good short-term payoffs for the cost of implementation - but that's not the assessment we are looking for here.

I interpret Anne's comments as suggesting that there might be more work involved than the proposal contemplates, that's all.  Some of the open issues are negated by detaching from the HTTP header grammar, and a whole new tranche are added.  I haven't reviewed the details there, but we'd need to be careful about quoting rules and the like.

This could be more opinionated about what it deems to be valuable and consider limiting this to writing of `__Host-`-prefixed names only, and never passing cookies that aren't `Secure`.  I'd like to have a discussion about that sort of question before we started on it.

I'm `worth prototyping` on this, with the recognition that there are probably other things that we should be investing effort in ahead of that.

annevk:
I'm concerned about the details. As there's still issues with `document.cookie` it's unclear that this new API got the architecture right (as such architecture hasn't really been defined to layer things on top on).

[...]

PR #1184, opened 2025-02-27
CookieStore for ServiceWorkers (bug 1947894)

bakulf:
For web-compat reasons,  we decided to implement CookieStoreManager. This commit updates our standard position about CookieStore to cover this ServiceWorker part.

### DIFF
--- a/activities.yml
+++ b/activities.yml
@@ -493,7 +493,7 @@ Cookie Store API:
   issue: 94
   position: positive
   rationale: |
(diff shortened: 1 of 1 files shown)


zcorpan (approved):

tantek (approved):
Good update to rationale. Our own implementation plans/concerns should be in Bugzilla or a spec issue rather than s-p.

zcorpan:
Linking the relevant issue for discoverability: #94

PR #1190, opened 2025-03-05
Positive on Declarative Web Push

### LINKED ISSUE #1176: [Push API] Declarative Web Push
### Specification title

Push API

### Specification or proposal URL (if available)

https://github.com/w3c/push-api/issues/360

### Explainer URL (if available)

_No response_

### Proposal author(s)

@beidson @marcoscaceres @annevk

### MDN URL

_No response_

### Caniuse.com URL

_No response_

### Bugzilla URL

https://bugzilla.mozilla.org/show_bug.cgi?id=1946085

### Mozillians who can provide input

@saschanaz @martinthomson @jrconlin @asutherland

### WebKit standards-position

_No response_

### Other information

The proposal intends to let push subscription independent from service workers and removes the need of calling SW event listener, achieving separate storage management and battery efficiency.

The PR: https://github.com/w3c/push-api/pull/385
A split issue for `mutable`: https://github.com/w3c/push-api/issues/391

saschanaz:
Closes #1176

### DIFF
--- a/activities.yml
+++ b/activities.yml
@@ -588,7 +588,6 @@ Declarative Web Push:
     Declarative Web Push introduces a JSON format that the user agent can recognize and parse without a Service Worker.
   id: declarative-web-push
   issue: 1176
-  position: positive
   rationale: |
     Mozilla is happy with the overall idea of Declarative Web Push. We believe having a declarative way to subscribe to push notifications will help users by reducing power usage on mobile devices and also help web authors by reducing required code size. We also believe that this introduces a way for browsers to suppress potentially unwanted push notifications from tracking users by conditionally removing its ability to use local state for mutable notifications.
 
@@ -599,8 +598,6 @@ Declarative Web Push:
 
     Each notification would be rendered purely from the declarative data in that case, thus achieving tracking protection without deactivating the whole push subscription.
   url: https://github.com/w3c/push-api/pull/385
-  venues:
-  - W3C
 Default Accessibility Semantics for Custom Elements:
   description: |
     This will allow custom elements to have "default" accessibility semantics, analogous to how built-in elements have "implicit" or "native" semantics.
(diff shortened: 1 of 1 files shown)


activities.yml
```diff
@@ -582,6 +582,22 @@ Declarative Shadow DOM:
...
   venues:
   - WHATWG
+Declarative Web Push:
+  bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1946085
+  description: |
+    Declarative Web Push introduces a JSON format that the user agent can recognize and parse without a Service Worker.
+  id: declarative-web-push
+  issue: 1176
+  rationale: |
+    Mozilla is happy with the overall idea of Declarative Web Push. We believe having a declarative way to subscribe to push notifications will help users by reducing power usage on mobile devices and also help web authors by reducing required code size. We also believe that this introduces a way for browsers to suppress potentially unwanted push notifications from tracking users by conditionally removing its ability to use local state for mutable notifications.
```
martinthomson, line 592: So... I don't believe Apple's claims about power usage here.  It's obviously less power consumption to not run code, but I don't find it convincing that instantiating and running a small SW is significant relative to activating the device radio, the speaker/haptics, or other costs that come with notifications.

What I do think is that this model of notifications is probably better overall than what we originally specified.  Having a clear path to a notification, with the option to intercept and update application state, is useful.  Having the option be truly optional in the sense that the browser might choose not to invoke it, is also good.

I do think that this is an incomplete solution, in that having no declarative way to dismiss a notification using this mechanism is problematic.  Mostly from the perspective of not enabling the creation of annoyances.

As for the rest of the text in this chunk, I would suggest that you post the other stuff on the issue, rather than here.  It's enough to just say that we think this is good.

saschanaz: Actually do we need a PR at all or is adding a tag to the issue enough?

zcorpan: Commenting in the issue and adding a label is enough. PR is needed if we want to include a description/rationale in the dashboard.

zcorpan (approved):

PR #1210, opened 2025-04-23
Upgrade to ubuntu-latest in GH workflow

zcorpan:
ubuntu-20.04 is no longer supported.


tantek (approved):
Seems fine, doesn't impact python version

PR #1314, opened 2025-11-10
Disallow smart quotes in activities.yml

zcorpan:
I had a patch for this locally because the smart quotes broke something. I don't remember what it was that broke, though.


tantek (approved):
Looks like a reasonable update. Could split the activities and python bits into separate PRs but also seems ok to land both, and we can revert the python bits if there are any problems.

PR #1328, opened 2025-12-08
Remove Web Share API entry

zcorpan:
The position label was removed in #27


tantek (approved):
Agreed. When we've updated #27 then we can rewrite this accordingly. Removal is the consistent thing to do for now.

PR #1353, opened 2026-02-04
Add WebAssembly Phase 3 proposals entry

### LINKED ISSUE #1350: WebAssembly Phase 3 Proposals
### Specification title

WebAssembly

### Specification or proposal URL (if available)

https://webassembly.github.io/spec

### Explainer URL (if available)

_No response_

### Proposal author(s)

_No response_

### MDN URL

https://developer.mozilla.org/en-US/docs/WebAssembly

### Caniuse.com URL

https://caniuse.com/wasm

### Bugzilla URL

_No response_

### Mozillians who can provide input

_No response_

### WebKit standards-position

_No response_

### Other information

This is a similar request to #527 where we adopted a blanket positive position for all phase 3 TC39 proposals.

The WebAssembly CG (in the W3C) has a [phase process](https://github.com/WebAssembly/meetings/blob/main/process/phases.md) for standardization similar (but not identical) to TC39. The SpiderMonkey team reviews all the [active proposals](https://github.com/webassembly/proposals?tab=readme-ov-file) in the process, and we only allow proposals that we are positive on to advance to phase 3 (the implementation phase).

Generally most wasm proposals don't have an impact on the larger web platform (e.g. a new arithmetic instruction) and don't need wider review. It'd be nice to have a blanket policy for them. There are some though that have a platform impact and may need a dedicated standards-position.

eqrion:
Fixes #1350. This is a copy of the language for the TC39 version. I don't think it needs any adjustments, but happy to workshop it.

### DIFF
--- a/activities.yml
+++ b/activities.yml
@@ -1934,6 +1934,10 @@ WebAssembly JS Promise Integration:
   url: https://github.com/WebAssembly/js-promise-integration/
   venues:
   - Proposal
+WebAssembly Phase 3 Proposals:
+  issue: 1350
+  rationale: |
+    Due to the structure of the WebAssembly CG, Phase 3 signifies that a proposal is ready to implement. Phase 3 proposals have been reviewed by the SpiderMonkey team and more broadly within Gecko. Phase 2 proposals which require security/privacy review or host integration require their own Mozilla Standards Position.
 WebAssembly Relaxed SIMD:
   description: |
     New 'relaxed' SIMD instructions for WebAssembly that have local non-determinism that depends on the platform.
--- a/activities.yml
+++ b/activities.yml
@@ -1937,7 +1937,7 @@ WebAssembly JS Promise Integration:
 WebAssembly Phase 3 Proposals:
   issue: 1350
   rationale: |
-    Due to the structure of the WebAssembly CG, Phase 3 signifies that a proposal is ready to implement. Phase 3 proposals have been reviewed by the SpiderMonkey team and more broadly within Gecko. Phase 2 proposals which require security/privacy review or host integration require their own Mozilla Standards Position.
+    Due to our high level of engagement in the WebAssembly standards process and the maturity phases used, we do not provide positions for proposals at Phase 3 if the scope of the proposal is limited to the WebAssembly language. Proposals that might affect other parts of the platform, have security- or privacy-relevant aspects, and less mature proposals should all request a position.
 WebAssembly Relaxed SIMD:
   description: |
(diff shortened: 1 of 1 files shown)


activities.yml
```diff
@@ -1934,6 +1934,10 @@ WebAssembly JS Promise Integration:
   url: https://github.com/WebAssembly/js-promise-integration/
   venues:
   - Proposal
+WebAssembly Phase 3 Proposals:
+  issue: 1350
+  rationale: |
+    Due to the structure of the WebAssembly CG, Phase 3 signifies that a proposal is ready to implement. Phase 3 proposals have been reviewed by the SpiderMonkey team and more broadly within Gecko. Phase 2 proposals which require security/privacy review or host integration require their own Mozilla Standards Position.
```
martinthomson: @eqrion, is there a reason you condition the phase 2 review?  Can you describe a case where we wouldn't review something at phase 2?  (I assume that we'd be happy to take requests for things at phase 1 or earlier, or would that also be less good for other reasons?)

eqrion: Hmm, that wording is maybe a little unclear.

What I'd like is for this position to just apply to proposals that are:
  1. Scoped to just the wasm language without broader web platform impact (such as security/privacy/Intl/DOM etc). Things like [wide-arithmetic](https://github.com/WebAssembly/wide-arithmetic/blob/main/proposals/wide-arithmetic/Overview.md), but not [wasm csp](https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md).
  2. At least phase 3 which means that we have expressly voted that it's a good enough design to start implementing within the committee. [Phase 2](https://github.com/WebAssembly/meetings/blob/main/process/phases.md#1-feature-proposal-community-group) proposals are less mature. The bar to phase 2 is only 'reasonably high level of consensus'.

For the things that don't meet (1), they should still need to get a dedicated S-P. For things that don't meet (2), they don't need a dedicated S-P because they're just not mature enough.

I could maybe just drop the 'Phase 2' part and say "Proposals which require security/privacy review or host integration require their own Mozilla Standards Position".

martinthomson: Maybe I can suggest:

> Due to the maturity phases used for WebAssembly proposals, we do not provide positions for proposals at Phase 3 if the effect of the proposal is limited to WebAssembly.  Proposals that might affect other parts of the platform, have security- or privacy-relevant aspects, and less mature proposals should all request a position.

eqrion: I think that's better, I took that and tweaked it a bit.

tantek (approved):
LGTM. If @martinthomson is happy with the edits to the PR, then let's land this.

martinthomson (approved):

